PR #62, opened 2026-04-12
feat: add LinkButton component

### LINKED ISSUE #11: Create LinkButton Component
### Description

Image

- https://www.figma.com/design/itAzcC43MCVpSyJCcmQJgu/ReDi-Events?node-id=1-1422

Create a `LinkButton` component for navigation actions (e.g., "← GO BACK").

### Props

```typescript
interface LinkButtonProps {
  label: string;
  icon?: React.ReactNode;
}
```

### Acceptance Criteria

- [x] Renders as a text link with optional icon
- [x] Storybook stories

Aarbei:
- resolves #11 

Implements the LinkButton component according to the requirements.

The component renders a text link with an optional icon and is designed to be reusable and responsive.

Changes:
- Added LinkButton component;
- Implemented optional icon prop with conditional rendering;
- Added responsive Tailwind styles;
- Implemented hover state;
- Added Storybook stories (Default (with icon), WithoutIcon).


frontend/src/components/LinkButton/LinkButton.types.ts
DariaAleshina, line 1: Hi! To avoid declaring `href` manually you could extend `AnchorHTMLAttributes<HTMLAnchorElement> `instead of manually declaring it: 
```
export interface LinkButtonProps extends AnchorHTMLAttributes<HTMLAnchorElement> {
  label: string;
  icon?: React.ReactNode;
}
```
This gives you href, className, target, aria-* etc. for free. You will need to spread `...rest` onto the `<a>` tag to forward them.

frontend/src/components/LinkButton/LinkButton.tsx
```diff
@@ -0,0 +1,14 @@
+import type { LinkButtonProps } from './LinkButton.types';
+
+const LinkButton = ({ label, icon, href }: LinkButtonProps) => {
+  return (
+    <a
+      href={href}
+      className="flex items-center justify-center gap-1.5 md:gap-2.5 w-fit px-2 py-3 md:px-3 md:py-4 text-[10px] md:text-[16px] font-medium text-[#ff5710]"
```
DariaAleshina: We should move all styles to a separate file (.**styles.ts**) and use Tailwind Variants (tv). Also i see hardcoded colors, you should use design tockens from **global.css** (and i see it is already there as `--color-button-secondary: #ff5710;`. Same goes for font sizes. 

I also noticed paddings you used are quite big:

Image

In the Figma design component is quite compact, i would check this out with designer (in our case @fabiorodriguesdev): 

Image

DariaAleshina:
@Aarbei you mention you implemented hover state but i don't see it :(

frontend/src/components/LinkButton/LinkButton.tsx
fabiorodriguesdev: yes, the space of the button between other elements are not part of the button itself, so, no paddings/margins needed. just `w-fit` to make the size of it always be the same as the text inside

MaksVintskovsky: I'd like to add that font size  text-[10px] md:text-[16px] also hardcoded and doesn't match the layout

fabiorodriguesdev (approved):

PR #65, opened 2026-04-16
9 create userarea component

### LINKED ISSUE #9: Create UserArea Component
### Description

Create a `UserArea` component that shows the logged-in user's avatar, name, and a dropdown menu with actions (Profile, Sign Out).

### States

- **Collapsed:** shows avatar + name + chevron
- **Expanded:** shows dropdown with menu items

### Props

```typescript
interface UserAreaProps {
  userName: string;
  avatarInitials: string;
  onProfile: () => void;
  onSignOut: () => void;
}
```

### Acceptance Criteria

- [ ] Displays circular avatar with initials + user name + dropdown chevron
- [ ] Click toggles dropdown menu
- [ ] Dropdown includes "Profile" (with icon) and "Sign out" (with icon)
- [ ] Clicking outside closes the dropdown
- [ ] Storybook stories: collapsed, expanded
- [ ] Accessible: `aria-expanded`, `aria-haspopup`, keyboard support

wms198:
- resolves #9 

image

### DIFF
--- a/frontend/src/components/UserArea/UserArea.stories.tsx
+++ b/frontend/src/components/UserArea/UserArea.stories.tsx
@@ -1,4 +1,3 @@
-import React from 'react';
 import type { Meta, StoryObj } from '@storybook/react-vite';
 
 import UserArea from './UserArea';
--- a/frontend/src/components/UserArea/UserArea.stories.tsx
+++ b/frontend/src/components/UserArea/UserArea.stories.tsx
@@ -1,3 +1,4 @@
+import React from 'react';
 import type { Meta, StoryObj } from '@storybook/react-vite';
 
 import UserArea from './UserArea';
@@ -11,7 +12,30 @@ const meta: Meta<typeof UserArea> = {
 export default meta;
 type Story = StoryObj<typeof UserArea>;
 
-export const Default: Story = {
+export const Expanded: Story = {
+  render: (args) => {
+    //XXX: no better way to expand without changeing interface
+    React.useEffect(() => {
+      const interval = setInterval(() => {
+        const trigger = document.querySelector('[aria-haspopup="true"][aria-expanded="false"]');
+        if (trigger) {
+          trigger.click();
+          clearInterval(interval);
+        }
+      }, 50);
+
+      return () => clearInterval(interval);
+    }, []);
+
+    return <UserArea {...args} />;
+  },
+  args: {
+    userName: 'George Michel',
+    avatarInitials: 'GM',
+  },
+};
+
+export const Collapsed: Story = {
   args: {
     userName: 'Lucy Lu',
     avatarInitials: 'LL',
--- a/frontend/src/components/UserArea/UserArea.stories.tsx
+++ b/frontend/src/components/UserArea/UserArea.stories.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import type { Meta, StoryObj } from '@storybook/react-vite';
+import type { Meta, StoryObj } from '@storybook/nextjs-vite';
 import UserArea from './UserArea';
 
 const meta: Meta<typeof UserArea> = {
--- a/frontend/src/components/UserArea/UserArea.stories.tsx
+++ b/frontend/src/components/UserArea/UserArea.stories.tsx
@@ -13,25 +13,15 @@ export default meta;
 type Story = StoryObj<typeof UserArea>;
 
 export const Expanded: Story = {
-  render: (args) => {
-    //XXX: no better way to expand without changeing interface
-    React.useEffect(() => {
-      const interval = setInterval(() => {
-        const trigger = document.querySelector('[aria-haspopup="true"][aria-expanded="false"]');
-        if (trigger) {
-          trigger.click();
-          clearInterval(interval);
-        }
-      }, 50);
-
-      return () => clearInterval(interval);
-    }, []);
-
-    return <UserArea {...args} />;
-  },
   args: {
     userName: 'George Michel',
     avatarInitials: 'GM',
+  },
+
+  play: async ({ canvasElement }) => {
+    const trigger = canvasElement.querySelector('[aria-haspopup="menu"]') as HTMLElement;
+
+    trigger?.click();
   },
 };
 
--- a/frontend/src/components/UserArea/UserArea.stories.tsx
+++ b/frontend/src/components/UserArea/UserArea.stories.tsx
@@ -1,5 +1,5 @@
+import React from 'react';
 import type { Meta, StoryObj } from '@storybook/react-vite';
-
 import UserArea from './UserArea';
 
 const meta: Meta<typeof UserArea> = {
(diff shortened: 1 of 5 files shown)


frontend/src/components/UserArea/UserArea.stories.tsx
```diff
@@ -13,26 +13,16 @@ export default meta;
...
-
-    return <UserArea {...args} />;
-  },
   args: {
     userName: 'George Michel',
     avatarInitials: 'GM',
   },
+
+  play: async ({ canvasElement }) => {
+    const trigger = canvasElement.querySelector('[aria-haspopup="menu"]') as HTMLElement;
```
wms198: I think we should show dropdown list based on ticket requirement, so that is why I added [aria-expanded="false"]'

wms198:
- 9 create userarea component

DariaAleshina:
Hi! Nice job! It expands/collapses nicely, tho I noticed 

- the links/buttons don't have enough gap between each other vs Figma visual:

image

Figma:
image

- You use generic` <div> ` for interactive elements (buttons) which causes accessibility issues - they are not focusable and cannot be reached by a tab-key on a keyboard (as a result both keyboard and screen reader users basically cannot use this feature). You should use semantic `<button>` element. (P.S just noticed you used role="button" to the main trigger, so it is accessible but lost its focus style. You should either add it then, or just use `<button> `here for simplicity imho).

- It would be also a good practice to exclude all pure decorative svg icons from accessibility tree so that they are not pronounced to screenreader users. here is the example from what we did InputField component:
`<EyeIcon size={20} aria-hidden="true" focusable="false" />`

- i would add cursor pointer to the main trigger button

frontend/src/components/UserArea/UserArea.stories.tsx
```diff
@@ -0,0 +1,32 @@
...
+const meta: Meta<typeof UserArea> = {
+  title: 'Components/UserArea',
+  component: UserArea,
+  tags: ['autodocs'],
+};
+
+export default meta;
+type Story = StoryObj<typeof UserArea>;
+
+export const Expanded: Story = {
```
DariaAleshina, line 23: the Expanded view is not really expanded 😅

Image

frontend/src/components/UserArea/UserArea.tsx
```diff
@@ -0,0 +1,75 @@
...
+      document.body.removeEventListener('click', closeDropdown);
+      document.removeEventListener('keydown', handleEscape);
+    };
+  }, []);
+
+  return (
+    <div className="flex">
+      <div
+        ref={ref}
+        className={userAreaStyles.base.join(' ')}
```
DariaAleshina: To avoid using "base.join" everywhere you should check out slots in TV: https://www.tailwind-variants.org/docs/slots

frontend/src/components/UserArea/UserArea.styles.ts
```diff
@@ -0,0 +1,37 @@
...
+    'absolute',
+  ],
+};
+
+export const dropDownStyles = {
+  base: ['flex', 'items-center ', 'justify-end', 'cursor-pointer', 'hover:text-grey-600'],
+};
+
+export const iconStyles = {
+  base: ['ml-2'],
```
DariaAleshina: to avoid styling icon here all all, you could just add `gap-2` on the wrapper (which you made "flex" already)

frontend/src/components/UserArea/UserArea.stories.tsx
wms198: yes, you are right. I just checked again on my computer and we will fix it.

frontend/src/components/UserArea/UserArea.styles.ts
wms198: Cloud you please explain why using 'gap-2'is better than 'ml-2'? Because we are using 'flex'? And gap is for controlling gutters between grid and flexbox items? ( I enjoy working more on backend and build apis, so the knowledge of frontend is not good). Thanks

frontend/src/components/UserArea/UserArea.tsx
```diff
@@ -0,0 +1,75 @@
...
+} from './UserArea.styles';
+import { useState, useEffect, useRef } from 'react';
+
+const UserArea = ({ userName, avatarInitials, onProfile, onSignOut }: UserAreaProps) => {
+  const [isExpanded, setIsExpended] = useState(false);
+  const ref = useRef<HTMLDivElement | null>(null);
+
+  useEffect(() => {
+    const closeDropdown = (e: MouseEvent) => {
+      if (ref.current && !ref.current.contains(e.target as Node)) {
```
DariaAleshina: I feel like here could be a bug. So the `ref ` is the main trigger button. So if I click anywhere outside it, dropdown menu collapses. Meaning if click on opened dropdown menu itself it will also collapse, which i guess might not be the behaviour we want. If you click on buttons SighOut or Profile, but do they perform the action before collapsing? Even if they do, seems wrong that it would collapse if i click on dropdown menu smw e.g. on the border/padding area around buttons. What do you think?

frontend/src/components/UserArea/UserArea.tsx
wms198: I will check the link out later, thanks for your share.

frontend/src/components/UserArea/UserArea.styles.ts
DariaAleshina: gap is for defining distance between flex items (or in grid  – between grid items) and this is exactly what you need here. you just define it on a "flex" parent with one style declaration and don't need bother child elements for it (imaging having more items than just two). Using margins works of course, but i would say its old school.

frontend/src/components/UserArea/UserArea.tsx
wms198: Yes, you are right. There is a bug, we will fix it.

Aarbei: Thanks for noticing. I have tried to solve this problem, it would be nice, if you take a look:)

DariaAleshina: Yes i looks great! 🤗🔥

frontend/src/components/UserArea/UserArea.tsx
```diff
@@ -0,0 +1,85 @@
+import { UserAreaProps } from './UserArea.types';
+import { Avatar } from '../Avatar';
+import { CaretDownIcon, UserIcon, SignOutIcon } from '@phosphor-icons/react';
+import { userArea } from './UserArea.styles';
+import { useState, useEffect, useRef } from 'react';
+
+const UserArea = ({ userName, avatarInitials, onProfile, onSignOut }: UserAreaProps) => {
+  const [isExpanded, setIsExpended] = useState(false);
```
ssda-developer: I noticed a small typo in the state names: it's currently a mix of isExp(a)nded and setIsExp(e)nded.

frontend/src/components/UserArea/UserArea.tsx
```diff
@@ -0,0 +1,85 @@
...
+    document.addEventListener('keydown', handleEscape);
+
+    return () => {
+      document.removeEventListener('pointerdown', closeDropdown, true);
+      document.removeEventListener('keydown', handleEscape);
+    };
+  }, []);
+
+  return (
+    <div className="relative inline-flex">
```
ssda-developer: Should we move the inline classes to UserArea.styles.ts to keep the styling consistent with the rest of the component?

frontend/src/components/UserArea/UserArea.tsx
wms198: When you clicked on the dropdown, did it stay? I just tried, but it collapsed.
But it might be my issue. When I rebased, I also got some warning that I skipped a few commit.

frontend/src/components/UserArea/UserArea.styles.ts
```diff
@@ -0,0 +1,37 @@
+export const userAreaStyles = {
+  base: [
+    'text-text-secondary',
+    'flex',
+    'justify-between',
+    'items-center',
+    'p-2',
+    'rounded-full',
+    'hover:bg-grey-100',
```
wms198: Hi, Aliona. I think mouse hover does not work. I dont know really why, but I am trying as well.

Image

 As you can see that I added bg color on button, the color doesn't change.

frontend/src/components/UserArea/UserArea.tsx
Aarbei: Thanks! We will fix it:)

frontend/src/components/UserArea/UserArea.tsx
Aarbei: Yes! It should be moved, thanks.

frontend/src/components/UserArea/UserArea.tsx
DariaAleshina: @wms198 for me it collapses now only when i click on a link (Profile or Signout), which is now intentional taken the code (added to onClick). We could also argue that maybe it should not collapse on link/button click, i am just not so sure what is the right way exactly.

fabiorodriguesdev (approved):
I'm approving and merging this PR since it's been open for too long. Component looks good.

PR #69, opened 2026-04-20
37 create eventcard component

santi-b-b:
### Changes
- **Component**: EventCard with props for event details (date, title, author, description, attendee count)
- **Types**: Comprehensive TypeScript types with JSDoc documentation for all props
  - `EventCardProps` — Union type supporting loading and loaded states
  - `EventCardAction` — Action type ('join' | 'leave' | 'edit')
  - Conditional prop rendering based on `isLoading` state
- **Stories**: Storybook stories demonstrating all variants and states
- **Styling**: Tailwind CSS with responsive design and hover effects
- **Documentation**: JSDoc comments on all props with default values

### Features
- ✅ Loading skeleton state support
- ✅ Dynamic action button with variant mapping (join → positive, leave → negative, edit → idle)
- ✅ Attendee count display (current/max)
- ✅ Interactive Card with hover effects
- ✅ Fully typed with TypeScript
- ✅ Storybook integration

### Files Added
- `src/components/EventCard/EventCard.tsx`
- `src/components/EventCard/EventCard.types.ts`
- `src/components/EventCard/EventCard.styles.ts`
- `src/components/EventCard/EventCardSkeleton.tsx`
- `src/components/EventCard/EventCard.stories.tsx`

### Testing
- Verify all story variants load correctly in Storybook
- Test with and without `isLoading` prop
- Verify action button text changes based on action type


frontend/src/components/EventCard/EventCard.tsx
```diff
@@ -0,0 +1,48 @@
...
+        </div>
+        <div className="flex justify-between">
+          <div className="flex gap-2.5 items-center text-input-secondary">
+            <UsersIcon size={20} />
+            <p className="text-sm">
+              {props.attendeeCount} of {props.maxAttendees}
+            </p>
+          </div>
+          <Button variant={buttonVariant} size="small" onClick={props.onActionClick}>
+            <p className="uppercase">{props.action}</p>
```
DariaAleshina: Hi! Seems like an over engineered construction. I don't think you need `<p> `here at all. It should work fine as well with `<Button>{props.action}</Button> `and add `uppercase` directly to buttonVariant styles.

frontend/src/components/EventCard/EventCard.tsx
```diff
@@ -0,0 +1,48 @@
...
+        </div>
+        <div className="flex flex-col gap-7.5">
+          <div>
+            <p className="text-xl">{props.title}</p>
+            <p className="text-sm text-ec-tertiary">{props.author}</p>
+          </div>
+          <p className="text-base text-ec-secondary">{props.description}</p>
+        </div>
+        <div className="flex justify-between">
+          <div className="flex gap-2.5 items-center text-input-secondary">
```
DariaAleshina: 🙂All styles should be moved to styles.tsx. check out slots for multiple elements: https://www.tailwind-variants.org/docs/slots. I used it in InputField component, you can check out for reference.

santi-b-b: thanks daria, I’ll check it out 🙂

frontend/src/components/EventCard/EventCard.tsx
santi-b-b: hey thanks for the comment. I did it this way because I’m not entirely sure every button will have an uppercase text. If that will be the case for the whole project, I agree we can pass it to the button styles.

frontend/src/components/EventCard/EventCard.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+export default function EventCard(props: EventCardProps) {
+  if (props.isLoading) {
+    return <EventCardSkeleton />;
+  }
+
+  const buttonVariant = variantMap[props.action];
+
+  return (
+    <Card interactive>
```
DariaAleshina, line 37: I guess I am a bit confused here. `<Card interactive>` means that `interactive` is set to true, which also means that EventCard is permanently "elevated" as it does not depend on the prop at all.... also if i check the storybook toggling "elevated" does do change anything...in both cases on card hover i see border change. but maybe i did not understand the point of "elevated" prop🥹, let me know.

frontend/src/components/EventCard/EventCard.tsx
```diff
@@ -0,0 +1,48 @@
...
+      <div className="flex flex-col gap-7.5 font-sans">
+        <div className="flex gap-2.5 items-center text-[14px] h-6 text-input-primary">
```
DariaAleshina: `text-[14px] `should not be hardcoded

frontend/src/components/EventCard/EventCard.tsx
```diff
@@ -0,0 +1,48 @@
...
+import { Card } from '../Card';
+import { CalendarDotsIcon, UsersIcon } from '@phosphor-icons/react';
+import { Button } from '../Button';
+import EventCardSkeleton from './EventCardSkeleton';
+
+const variantMap = {
+  join: 'positive',
+  leave: 'negative',
+  edit: 'idle',
+} as const;
```
DariaAleshina, line 13: Nice! 🙂

frontend/src/components/EventCard/EventCard.tsx
santi-b-b: I actually took the props from the ticket and think I didn’t used the elevated in the end. 
interactive means that the change at hover, that does not happened on the loading case.

frontend/src/components/EventCard/EventCard.tsx
santi-b-b: 😉 I’ll include it in the css

frontend/src/components/EventCard/EventCard.tsx
santi-b-b: I’ll take it out 🙂

frontend/src/components/EventCard/EventCard.types.ts
```diff
@@ -0,0 +1,73 @@
...
+   * Whether the card is in loading state.
+   * @default false
+   */
+  isLoading?: false;
+
+  /**
+   * Date of the event in the format: April 4, 2017 – 2:17 PM
+   * @default 'April 4, 2017 – 2:17 PM'
+   */
+  date: string;
```
DariaAleshina: @fabiorodriguesdev (just out of curiosity) do i understand correctly that we will need to transform timestamp from db into this specific string format ('April 4, 2017 – 2:17 PM') before passing to a component?

frontend/src/components/EventCard/EventCard.stories.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+export default meta;
+type Story = StoryObj<typeof EventCard>;
+
+export const EventCardExample: Story = {
+  args: {
+    isLoading: false,
+    elevated: false,
+    action: 'join',
+    date: 'April 4, 2017 – 2:17 PM',
```
DariaAleshina: I think you need to change arg type type for the date, as in this case it should be string, but the storybook offers a "date/time" input type (see the right colunm):

Image

And as soon as i choose a custom date, it displays raw timestamp:

Image

santi-b-b: Let’s wait for fabio’s answer to you previous comment. I think what’s happening here is that storybook thinks it is a date type because of the prop name. but at the time is not, is a string. 🤔

frontend/src/components/EventCard/EventCard.stories.tsx
```diff
@@ -0,0 +1,78 @@
...
+};
+
+export default meta;
+type Story = StoryObj<typeof EventCard>;
+
+export const EventCardExample: Story = {
+  args: {
+    isLoading: false,
+    elevated: false,
+    action: 'join',
```
DariaAleshina, line 25: i would recommend for storybook for "join" prop specifically to define arg type options so that others could test easily. right now by default it could be any string here: 

Image

but as this prop has specific options, you could do smth sililar as i did for "radius" and "animation" props: 
```
const meta: Meta<typeof Skeleton> = {
  title: 'Components/Skeleton',
  component: Skeleton,
  argTypes: {
    animation: {
      options: ['wave', 'pulse', false],
      control: {
        type: 'radio',
      },
    },
    radius: {
      options: ['none', 'sm', 'base', 'md', 'lg', 'full'],
      control: {
        type: 'radio',
      },
    },
  },
  tags: ['autodocs'],
};
```

which in storybook will allow to test easier: 

Image

DariaAleshina:
Hi @santi-b-b ! Overall 🙂good job – visually component looks exactly as in Figma file. Still I left some comments for the code itself. If something is not clear, DM me in Slack :)

frontend/src/components/EventCard/EventCard.tsx
DariaAleshina: Yeah you are right, i guess we should assume that uppercase is not the only option here. Then i would suggest to change `<p>` to `<span>` as 1) it would be more semantically correct as button label is not really a paragraph and 2) span is a generic inline element which is exactly what you need for pure styling purposes.

frontend/src/components/EventCard/EventCard.stories.tsx
DariaAleshina: you can override the arg types, probably smth like this: 

```
const meta: Meta<typeof EventCard> = {
  title: 'Components/EventCard',
  component: EventCard,
  tags: ['autodocs'],
  argTypes: {
    date: { control: 'text' },
  },
};
```

frontend/src/components/EventCard/EventCard.types.ts
fabiorodriguesdev: We don't need to transform in the component itself BUT is also not a bad idea.
Another point is: @default only necessary for optional props, if is mandatory you shouldn't pass anything by default so the user sees the error message in the IDE that something is missing.

frontend/src/components/EventCard/EventCard.tsx
fabiorodriguesdev: Nice indeed!

fabiorodriguesdev (commented):
Good job overall!

frontend/src/components/EventCard/EventCard.tsx
santi-b-b: I changed it to span, thanks for the comment : )

frontend/src/components/EventCard/EventCard.tsx
santi-b-b: Captura de pantalla 2026-04-21 143619


Added to the styles file : ) thanks for the suggestion

frontend/src/components/EventCard/EventCard.tsx
santi-b-b: thanks 😄

santi-b-b:
@DariaAleshina & @fabiorodriguesdev thanks for all the comments I just committed the changes that I think take all the suggestions. Please take a look at them whenever you have the time : )

- moved the component styles to the styles file.
- took out the unused elevated prop.
- changed the hardcoded text-14px for text-sm
- change to type of the date prop from string to date and gave it the format inside the component.
- change the argtype of the action prop for easier testing in storybook
- took out the default values for mandatory properties

frontend/src/components/EventCard/EventCard.tsx
```diff
@@ -0,0 +1,75 @@
...
+  } = EventCardStyles();
+
+  const buttonVariant = variantMap[props.action];
+
+  const datePart = new Intl.DateTimeFormat('en-US', {
+    month: 'long',
+    day: 'numeric',
+    year: 'numeric',
+  }).format(props.date);
+
```
freeediver: It would make more sense to move this part of date formatting inside the utilities file for the re-usability by other components which will need it.

frontend/src/components/EventCard/EventCard.tsx
```diff
@@ -0,0 +1,75 @@
...
+  const timePart = new Intl.DateTimeFormat('en-US', {
+    hour: 'numeric',
+    minute: '2-digit',
+    hour12: true,
+  }).format(props.date);
+
```
freeediver: same too for timePart, it should be centralized inside the utility file.

santi-b-b: hey where is the utility file? I can't find it 😅

santi-b-b: it is now a function on the utils file 😃 thanks for the suggestion

frontend/src/components/EventCard/EventCard.types.ts
```diff
@@ -0,0 +1,66 @@
...
+
+type EventCardLoadedProps = EventCardBaseProps & {
+  /**
+   * Whether the card is in loading state.
+   * @default false
+   */
+  isLoading?: false;
+
+  /**
+   * Date of the event in stardart ISO: 2017-04-04T14:17:00Z
```
DariaAleshina, line 30: are you sure its gonna be in ISO format? we could be getting timestamp integer format, no? Could be confusing if so

frontend/src/components/EventCard/EventCard.types.ts
```diff
@@ -0,0 +1,66 @@
...
+   */
+  date: Date;
+
+  /**
+   * Title of the event.
+   * @default 'How to Network'
```
DariaAleshina: There is no default for title, author, description, attendees number (maybe smth else i missed), yet the comments in docs are confusing

frontend/src/components/EventCard/EventCardSkeleton.tsx
```diff
@@ -0,0 +1,32 @@
+import { Card } from '../Card';
+import { Skeleton } from '../Skeleton';
+
+export default function EventCardSkeleton() {
+  return (
+    <Card>
+      <div className="flex flex-col gap-7.5">
```
DariaAleshina: i guess the skeleton could use styles definded in styles.ts as well 😉 (esp. if sticking to the rule of moving styles to a separate file)

frontend/src/utils/utils.ts
```diff
@@ -15,3 +15,19 @@ export const getInitials = (name: string): string =>
...
+  }).format(date);
+
+  const timePart = new Intl.DateTimeFormat('en-US', {
+    hour: 'numeric',
+    minute: '2-digit',
+    hour12: true,
+  }).format(date);
+
+  return `${datePart} – ${timePart}`;
+};
```
DariaAleshina, line 33: Cool!

frontend/src/components/EventCard/EventCard.types.ts
santi-b-b: from what I read if it was the most probable format coming from express. what do you think if we keep it this way and once we have a definition on how it is coming from the backend we can adjust it?

DariaAleshina: ahh ok, thanx for the insight, then lets leave as is

DariaAleshina (approved):

PR #72, opened 2026-04-21
feat: update User model with auth fields and Role enum

### LINKED ISSUE #29: Update user data model
Right now the User model only stores `name` and `email`. To support login, we need to store a hashed password and know whether a user is a regular user or an admin.
 
### Implementation Notes
 
Update `prisma/schema.prisma`:
 
```prisma
model User {
  id           String   @id @default(uuid())
  email        String   @unique
  firstName    String
  lastName     String
  passwordHash String
  role         Role     @default(USER)
  createdAt    DateTime @default(now())
  updatedAt    DateTime @updatedAt
}
 
enum Role {
  USER
  ADMIN
}
```
 
Then sync the database and regenerate the client:
 
```bash
npm run prisma:push
npm run prisma:generate
```
 
> Never store a plain password. `passwordHash` will hold the result of running the password through `bcrypt` (done in a later task).
 
### Acceptance Criteria
 
- [ ] `passwordHash` and `role` fields exist on the User model
- [ ] `firstName` and `lastName` replace the old `name` field
- [ ] `Role` enum exists with `USER` and `ADMIN` values
- [ ] `npm run prisma:studio` shows the new fields in the User table

akhilachiju:
- resolves #29

Update User model with auth fields and Role enum

What changed
- Replaced name field with firstName and lastName
- Added passwordHash field to store bcrypt-hashed passwords
- Added role field with Role enum (USER / ADMIN), defaults to USER
- Updated userService.ts to match the new schema fields

Why
To support user authentication in upcoming tasks. Passwords will never be stored in plain text — passwordHash will hold the bcrypt result.

Steps to apply
bash
npm run prisma:push
npm run prisma:generate

### DIFF
--- a/backend/prisma/schema.prisma
+++ b/backend/prisma/schema.prisma
@@ -11,9 +11,17 @@ datasource db {
 }
 
 model User {
-  id        String   @id @default(uuid())
-  email     String   @unique
-  name      String?
-  createdAt DateTime @default(now())
-  updatedAt DateTime @updatedAt
+  id           String   @id @default(uuid())
+  email        String   @unique
+  firstName    String
+  lastName     String
+  passwordHash String
+  role         Role     @default(USER)
+  createdAt    DateTime @default(now())
+  updatedAt    DateTime @updatedAt
+}
+
+enum Role {
+  USER
+  ADMIN
 }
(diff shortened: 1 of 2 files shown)


fabiorodriguesdev (approved):
Good job!

PR #74, opened 2026-04-23
add error handling middleware

DariaAleshina:
- closes #43 
- Added **notFound** and **errorHandler** middleware in backend/src/middleware/errorHandler.ts
- Registered both middleware at the end of the Express app in index.ts (after all routes)
- Added `express-async-errors` to patch Express 4's missing async error propagation. Async errors thrown in controllers are now automatically forwarded to the error handler middleware, matching Express 5 behavior.

### DIFF
--- a/backend/src/middleware/errorHandler.ts
+++ b/backend/src/middleware/errorHandler.ts
@@ -8,23 +8,12 @@ export function notFound(req: Request, res: Response) {
 // Removing `next` silently breaks error handling — Express would treat this as regular middleware instead.
 
 export function errorHandler(
-  err: Error & {
-    code?: string;
-  },
+  err: Error,
   req: Request,
   res: Response,
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   next: NextFunction
 ) {
   console.error('Error:', err);
-
-  if (err.code === 'P2025') {
-    return res.status(404).json({ error: 'Record not found' });
-  }
-
-  if (err.code === 'P2002') {
-    return res.status(409).json({ error: 'Duplicate value' });
-  }
-
   return res.status(500).json({ error: 'Something went wrong' });
 }
--- a/backend/src/middleware/errorHandler.ts
+++ b/backend/src/middleware/errorHandler.ts
@@ -4,6 +4,9 @@ export function notFound(req: Request, res: Response) {
   res.status(404).json({ error: 'Page not found' });
 }
 
+// ESLint ignore required: Express detects error-handling middleware by its 4-parameter signature (err, req, res, next).
+// Removing `next` silently breaks error handling — Express would treat this as regular middleware instead.
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
 export function errorHandler(err: Error, req: Request, res: Response, _next: NextFunction) {
   console.error('Error:', err);
   res.status(500).json({ error: 'Something went wrong' });
--- a/backend/src/middleware/errorHandler.ts
+++ b/backend/src/middleware/errorHandler.ts
@@ -8,6 +8,6 @@ export function notFound(req: Request, res: Response) {
 // Removing `next` silently breaks error handling — Express would treat this as regular middleware instead.
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
 export function errorHandler(err: Error, req: Request, res: Response, next: NextFunction) {
-  console.error('Error:', err);
+  console.error('Error:', err.stack);
   return res.status(500).json({ error: 'Something went wrong' });
 }
--- a/backend/src/middleware/errorHandler.ts
+++ b/backend/src/middleware/errorHandler.ts
@@ -1,0 +1,10 @@
+import { NextFunction, Request, Response } from 'express';
+
+export function notFound(req: Request, res: Response) {
+  res.status(404).json({ error: 'Page not found' });
+}
+
+export function errorHandler(err: Error, req: Request, res: Response, next: NextFunction) {
+  console.error('Error:', err);
+  res.status(500).json({ error: 'Something went wrong' });
+}
--- a/backend/src/middleware/errorHandler.ts
+++ b/backend/src/middleware/errorHandler.ts
@@ -1,7 +1,7 @@
 import { NextFunction, Request, Response } from 'express';
 
 export function notFound(req: Request, res: Response) {
-  res.status(404).json({ error: 'Page not found' });
+  return res.status(404).json({ error: 'Page not found' });
 }
 
 // ESLint ignore required: Express detects error-handling middleware by its 4-parameter signature (err, req, res, next).
@@ -9,5 +9,5 @@ export function notFound(req: Request, res: Response) {
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
 export function errorHandler(err: Error, req: Request, res: Response, next: NextFunction) {
   console.error('Error:', err);
-  res.status(500).json({ error: 'Something went wrong' });
+  return res.status(500).json({ error: 'Something went wrong' });
 }
--- a/backend/src/middleware/errorHandler.ts
+++ b/backend/src/middleware/errorHandler.ts
@@ -7,7 +7,7 @@ export function notFound(req: Request, res: Response) {
 // ESLint ignore required: Express detects error-handling middleware by its 4-parameter signature (err, req, res, next).
 // Removing `next` silently breaks error handling — Express would treat this as regular middleware instead.
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
-export function errorHandler(err: Error, req: Request, res: Response, _next: NextFunction) {
+export function errorHandler(err: Error, req: Request, res: Response, next: NextFunction) {
   console.error('Error:', err);
   res.status(500).json({ error: 'Something went wrong' });
 }
--- a/backend/src/middleware/errorHandler.ts
+++ b/backend/src/middleware/errorHandler.ts
@@ -6,8 +6,25 @@ export function notFound(req: Request, res: Response) {
 
 // ESLint ignore required: Express detects error-handling middleware by its 4-parameter signature (err, req, res, next).
 // Removing `next` silently breaks error handling — Express would treat this as regular middleware instead.
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
-export function errorHandler(err: Error, req: Request, res: Response, next: NextFunction) {
-  console.error('Error:', err.stack);
+
+export function errorHandler(
+  err: Error & {
+    code?: string;
+  },
+  req: Request,
+  res: Response,
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  next: NextFunction
+) {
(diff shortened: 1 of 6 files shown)


backend/src/middleware/errorHandler.ts
```diff
@@ -0,0 +1,10 @@
+import { NextFunction, Request, Response } from 'express';
+
+export function notFound(req: Request, res: Response) {
+  res.status(404).json({ error: 'Page not found' });
+}
+
+export function errorHandler(err: Error, req: Request, res: Response, _next: NextFunction) {
```
Davies70: You're right. I tested `_` and it still triggers the lint error. Looks like the ESLint config needs to be updated to ignore it.

DariaAleshina: i added inline ignore as per Michael's reco :) it worked

backend/src/middleware/errorHandler.ts
```diff
@@ -0,0 +1,13 @@
+import { NextFunction, Request, Response } from 'express';
+
+export function notFound(req: Request, res: Response) {
+  res.status(404).json({ error: 'Page not found' });
```
freeediver: Missing return Statements.

Add `return` before `res.status()` in `errorHandler.ts` to prevent potential double responses.

Everything else looks solid. 👍

backend/src/middleware/errorHandler.ts
```diff
@@ -0,0 +1,13 @@
...
+}
+
+// ESLint ignore required: Express detects error-handling middleware by its 4-parameter signature (err, req, res, next).
+// Removing `next` silently breaks error handling — Express would treat this as regular middleware instead.
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+export function errorHandler(err: Error, req: Request, res: Response, next: NextFunction) {
+  console.error('Error:', err);
+  res.status(500).json({ error: 'Something went wrong' });
```
freeediver: here too.

freeediver:
Good job overall! One important thing to fix:  Controllers are bypassing the central error handle. 
- The problem: This defeats the purpose of having a central error handler. You're repeating error handling logic in every controller.
- What to fix: Remove the `try-catch` blocks from your controller methods. Let errors bubble up to `errorHandler` instead of sending `500` responses from each controller.
Here's an example :
```ts
async getUsers(req: Request, res: Response) {
  const users = await userService.getAllUsers();
  res.json({ users }); // ✅ Errors go to errorHandler
}
```

Alternative: If you need try-catch for specific cases, use next(error) to forward to the error handler.

DariaAleshina:
> Good job overall! One important thing to fix: Controllers are bypassing the central error handle.
> 
> * The problem: This defeats the purpose of having a central error handler. You're repeating error handling logic in every controller.
> * What to fix: Remove the `try-catch` blocks from your controller methods. Let errors bubble up to `errorHandler` instead of sending `500` responses from each controller.
>   Here's an example :
> 
> ```ts
> async getUsers(req: Request, res: Response) {
>   const users = await userService.getAllUsers();
>   res.json({ users }); // ✅ Errors go to errorHandler
> }
> ```
> 
> Alternative: If you need try-catch for specific cases, use next(error) to forward to the error handler.

@freeediver thanks for the comment! What you describe makes sense. So far I followed the ticket description which says the middleware should catch "any unhandled errors". @fabiorodriguesdev — shall we make `errorHandler` catch all errors, or keep it as a fallback only?

fabiorodriguesdev:
> > Good job overall! One important thing to fix: Controllers are bypassing the central error handle.
> > 
> > * The problem: This defeats the purpose of having a central error handler. You're repeating error handling logic in every controller.
> > * What to fix: Remove the `try-catch` blocks from your controller methods. Let errors bubble up to `errorHandler` instead of sending `500` responses from each controller.
> >   Here's an example :
> > 
> > ```ts
> > async getUsers(req: Request, res: Response) {
> >   const users = await userService.getAllUsers();
> >   res.json({ users }); // ✅ Errors go to errorHandler
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Alternative: If you need try-catch for specific cases, use next(error) to forward to the error handler.
> 
> @freeediver thanks for the comment! What you describe makes sense. So far I followed the ticket description which says the middleware should catch "any unhandled errors". @fabiorodriguesdev — shall we make `errorHandler` catch all errors, or keep it as a fallback only?

I would make this question to a teacher who is more specialized in Backend! please ask today during class @DariaAleshina

FrontierPsychiatrist:
Chiming in!

First of all I agree with @freeediver, the error handling middleware should replace the generic `try/catch` blocks in the controller that are responsible for converting any error in a `res.status(500)`. This will lose the specific error message in the response (like `{ error: 'Error fetching users' }` and replace it with the message from the middleware, but in my opinion these specific error messages do not add a lot for a caller of the endpoint. If I do `GET /users` and get a 500 I _know_ there is an error fetching users, so the error message does not help a lot.

Now comes the **but**:
Express version 4 does not play nice with `async` controller functions (i.e. `async foo(req, res)` or anything that returns a `Promise<...>`). See [this document](https://expressjs.com/en/guide/error-handling.html).

To test what happens _without_ any extra solution, try the following (with the error handling middleware): Remove the `try/catch` in one controller method and just throw an error explicitly

```typescript
async getUsers(req: Request, res: Response) {
  throw new Error('this is a test');
  const users = await userService.getAllUsers();
  res.json({ users });
}
```

Now start the application and invoke the get users endpoint. IIRC, the request will just hang forever, even with the error handling middleware from this PR.

Here are some solutions:
1. Wrap all handler functions with a non-error handling middleware that is Promise-aware.
    This is a solution that is a bit hard to understand in the beginning as it involves a higher-order function (i.e. a function that returns a function):
   ```typescript
   // Example in userRoutes.ts
   const userController = new UserController();
   const userRouter = Router();
   function asyncRoute(handler: (req: Request, res: Response) => Promise<unknown>) {
     return (req, res, next) => {
       handler(req, res).catch(next);
     };
   }
   userRouter.get('/', asyncRoute((req, res) => userController.getUsers(req, res)));
   ```
   Hope I didn't make a mistake here. The main issues with this are: It's hard to understand, and everyone (!) needs to remember to wrap all (async) route definitions with the `asyncRoute` helper.
2. Upgrade to Express 5. To my understanding, Express 5 works out of the box with async route handlers. I have never used it though and I can't currently anticipate how many extra changes this incurs.

If you remove the try/catch (which I really think you should - otherwise the generic error handling middleware isn't even needed really), then you need to do either 1 or 2.

Final note: I didn't try this out. I might be mistaken, especially about the initial test. I'd appreciate if you do the initial test (with the `throw new Error`) to see if there really is an issue with Express 4 here.

FrontierPsychiatrist:
The try/catch in userController still need to be removed, I think it would be matching to do this in this PR.

DariaAleshina:
> Chiming in!
> 
> First of all I agree with @freeediver, the error handling middleware should replace the generic `try/catch` blocks in the controller that are responsible for converting any error in a `res.status(500)`. This will lose the specific error message in the response (like `{ error: 'Error fetching users' }` and replace it with the message from the middleware, but in my opinion these specific error messages do not add a lot for a caller of the endpoint. If I do `GET /users` and get a 500 I _know_ there is an error fetching users, so the error message does not help a lot.
> 
> Now comes the **but**: Express version 4 does not play nice with `async` controller functions (i.e. `async foo(req, res)` or anything that returns a `Promise<...>`). See [this document](https://expressjs.com/en/guide/error-handling.html).
> 
> To test what happens _without_ any extra solution, try the following (with the error handling middleware): Remove the `try/catch` in one controller method and just throw an error explicitly
> 
> ```ts
> async getUsers(req: Request, res: Response) {
>   throw new Error('this is a test');
>   const users = await userService.getAllUsers();
>   res.json({ users });
> }
> ```
> 
> Now start the application and invoke the get users endpoint. IIRC, the request will just hang forever, even with the error handling middleware from this PR.
> 
> Here are some solutions:
> 
> 1. Wrap all handler functions with a non-error handling middleware that is Promise-aware.
>    This is a solution that is a bit hard to understand in the beginning as it involves a higher-order function (i.e. a function that returns a function):
>    ```ts
>    // Example in userRoutes.ts
>    const userController = new UserController();
>    const userRouter = Router();
>    function asyncRoute(handler: (req: Request, res: Response) => Promise<unknown>) {
>      return (req, res, next) => {
>        handler(req, res).catch(next);
>      };
>    }
>    userRouter.get('/', asyncRoute((req, res) => userController.getUsers(req, res)));
>    ```
>    
>    
>        
>          
>        
>    
>          
>        
>    
>        
>      
>    Hope I didn't make a mistake here. The main issues with this are: It's hard to understand, and everyone (!) needs to remember to wrap all (async) route definitions with the `asyncRoute` helper.
> 2. Upgrade to Express 5. To my understanding, Express 5 works out of the box with async route handlers. I have never used it though and I can't currently anticipate how many extra changes this incurs.
> 
> If you remove the try/catch (which I really think you should - otherwise the generic error handling middleware isn't even needed really), then you need to do either 1 or 2.
> 
> Final note: I didn't try this out. I might be mistaken, especially about the initial test. I'd appreciate if you do the initial test (with the `throw new Error`) to see if there really is an issue with Express 4 here.

Thank you @FrontierPsychiatrist you are right, i tested and async errors were not handled indeed. I reviewed possible solutions and decided to add the package that matches Express 5 behavior of automatically forwarding async err to the middleware. This way code is cleaner and we don't need to upgrade to Express 5. 

So try/catch in not required in controllers anymore and could be deleted at all. Yet I have not touched the existing code in controllers at this point as not part of this ticket. 
Could you review my PR again, approve it or add specific comments to improve middleware?

FrontierPsychiatrist:
@DariaAleshina I would see removal of try/catch as part of this PR. It doesn't need to be in the ticket - the ticket writer doesn't necessarily know that try/catch was used before. The move to a generic error handling includes the migration of old, now unnecessary code.

backend/src/middleware/errorHandler.ts
```diff
@@ -0,0 +1,30 @@
...
+    code?: string;
+  },
+  req: Request,
+  res: Response,
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  next: NextFunction
+) {
+  console.error('Error:', err);
+
+  if (err.code === 'P2025') {
```
FrontierPsychiatrist: This is what I would _not_ really recommend and what I talked about in the session on Tuesday. This is much too specific for a generic error handler. This should be handled at repository/service level, thrown to controller and then explicitly there.

You could argue that what I wrote still can be done and this is just a last resort I guess.

FrontierPsychiatrist (changes requested):
Opinion, but I'd remove the error handlers for some "codes" from the error middleware. It's too specific for a generic last-resort error handler.

backend/src/middleware/errorHandler.ts
DariaAleshina: Ok got it!

FrontierPsychiatrist (approved):

PR #76, opened 2026-04-23
feat: add NetworkError view and page

### LINKED ISSUE #16: Create Network Error Page
### Description

Image

- Desktop: https://www.figma.com/design/itAzcC43MCVpSyJCcmQJgu/ReDi-Events?node-id=0-1
- Mobile: https://www.figma.com/design/itAzcC43MCVpSyJCcmQJgu/ReDi-Events?node-id=1-869

Create a `NetworkErrorPage` using `PublicLayout`. Displayed when the app encounters a server or network error. Shows a friendly error message with a refresh action.

### Page Structure

```
PublicLayout (no nav link)
└── Content Container (centered in right column)
    ├── Heading: "Something went wrong"
    ├── Subheading: "Seems like we are having trouble in our servers, please try to refresh the page."
    └── Button (intent="refresh", label="REFRESH")
```

### Acceptance Criteria

- [ ] Uses `PublicLayout` without top-right nav link
- [ ] Heading "Something went wrong"
- [ ] Subheading with server trouble message
- [ ] Orange "REFRESH" button reloads the page
- [ ] Content vertically centered in the right column

ssda-developer:
- Closes #16 

@fabiorodriguesdev Hey! I decided to split the NetworkError page into three files: page.tsx (for Metadata), NetworkErrorView.tsx (client-side logic), and NetworkError.styles.ts (styles).

I’m not sure if this is our preferred pattern for simple pages. Is this modular approach okay, or should I combine them into a single file to keep it simpler?


frontend/src/app/(public)/network-error/NetworkErrorView.tsx
```diff
@@ -0,0 +1,28 @@
...
+        </p>
+      </div>
+      <Button variant="secondary" onClick={handleRefresh}>
+        REFRESH
+      </Button>
+    </div>
+  );
+};
+
+export default NetworkErrorView;
```
fabiorodriguesdev, line 28: Looks good, I would just prefer to use NextJS proper function for refresh:


```suggestion
'use client';

import { useRouter } from 'next/navigation'
import { Button } from '../../../components';
import { networkErrorStyles } from './NetworkError.styles';

const NetworkErrorView = () => {
  const styles = networkErrorStyles();

  const handleRefresh = () => {
    router.refresh()
  };

  return (
    <div className={styles.container()}>
      <div className={styles.textWrapper()}>
        <h1 className={styles.title()}>Something went wrong</h1>
        <p className={styles.description()}>
          Seems like we are having trouble in our servers, please try to refresh the page.
        </p>
      </div>
      <Button variant="secondary" onClick={handleRefresh}>
        REFRESH
      </Button>
    </div>
  );
};

export default NetworkErrorView;
```

fabiorodriguesdev:
The question about the component + styles, I personally don't have a preference for pages, whatever works fine for the others in the class is fine. I personally usually leave the styling of pages directly in the html since they don't carry any kind of props that will change how things looks. or if there are way too many tailwind classes then I separate, otherwise I rather keep simpler. but both ways are fine imo

ssda-developer:
> The question about the component + styles, I personally don't have a preference for pages, whatever works fine for the others in the class is fine. I personally usually leave the styling of pages directly in the html since they don't carry any kind of props that will change how things looks. or if there are way too many tailwind classes then I separate, otherwise I rather keep simpler. but both ways are fine imo

You're right, this page is quite small and static, so I've merged the styles into a single file. I also checked our not-found.tsx page and noticed it follows a similar pattern, so it makes sense to keep it consistent across the project.

frontend/src/app/(public)/network-error/NetworkErrorView.tsx
ssda-developer: I was actually debating between router.refresh() and window.location.reload(). I initially thought a full reload might be safer in case of connection loss, but I agree that router.refresh() is more efficient and provides a better UX. I’ve updated the code to use the Next.js router, it's definitely a much better approach here.

fabiorodriguesdev (approved):
👏🏼

PR #77, opened 2026-04-23
41 implement login endpoint post authlogin

MaksVintskovsky:
- resolves #41

### DIFF
--- a/backend/package.json
+++ b/backend/package.json
@@ -12,7 +12,7 @@
     "url": "git@github.com:ReDI-School/S26-Full-Stack-Circle.git"
   },
   "scripts": {
-    "dev": "prisma generate && tsx watch src/index.ts",
+    "dev": "prisma generate && tsx --require=dotenv/config --watch src/index.ts",
     "build": "prisma generate && tsc && tsc-alias",
     "start": "node dist/src/index.js",
     "lint": "eslint src/ prisma/seed.ts",
--- a/backend/package.json
+++ b/backend/package.json
@@ -35,8 +35,9 @@
     "cors": "^2.8.5",
     "dotenv": "^16.5.0",
     "express": "^4.18.2",
-    "helmet": "^8.1.0",
     "jsonwebtoken": "^9.0.3"
+    "express-async-errors": "^3.1.1",
+    "helmet": "^8.1.0"
   },
   "devDependencies": {
     "@eslint/js": "^10.0.1",
--- a/backend/package.json
+++ b/backend/package.json
@@ -31,14 +31,18 @@
     "@prisma/adapter-pg": "^7.5.0",
     "@prisma/client": "^7.5.0",
     "@types/cors": "^2.8.19",
+    "bcrypt": "^6.0.0",
     "cors": "^2.8.5",
     "dotenv": "^16.5.0",
     "express": "^4.18.2",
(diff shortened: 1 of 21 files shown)


backend/prisma/seed.ts
```diff
@@ -1,4 +1,40 @@
...
+
+// main()
+//   .catch((e) => {
+//     console.error('Error during seed:', e);
+//     process.exit(1);
+//   })
+//   .finally(async () => {
+//     await prisma.$disconnect();
+//   });
+
```
fabiorodriguesdev: if it's working, we don't need to keep commented code :)

backend/src/controllers/authController.ts
```diff
@@ -0,0 +1,36 @@
+import { Request, Response } from 'express';
+import { AuthService } from '../services/authService.js';
+
+export class AuthController {
+  private authService = new AuthService();
```
fabiorodriguesdev:
```suggestion
  private readonly authService = new AuthService();
```

Since `authService` is assigned once at declaration and never reassigned anywhere in the class, TypeScript/ESLint flags it as a candidate for `readonly`. Marking it `readonly` enforces at compile time that the property can't be reassigned after initialization, making the code's intent clearer and preventing accidental mutations.

fabiorodriguesdev (commented):
Good work!

backend/src/controllers/authController.ts
MaksVintskovsky: Thanks for review! Just to clarify: should I keep this PR unmerged for now and wait for the Sign In and Registration pages, so I can adapt the auth flow with them?

fabiorodriguesdev: do you think the auth flow will need refactor after the other tickets? I imagine it wont be necessary. I would merge it and if we see the need we can open a new Pr to adapt, what do you think?

backend/package.json
```diff
@@ -31,15 +31,19 @@
     "@prisma/adapter-pg": "^7.5.0",
     "@prisma/client": "^7.5.0",
     "@types/cors": "^2.8.19",
+    "bcrypt": "^6.0.0",
     "cors": "^2.8.5",
     "dotenv": "^16.5.0",
     "express": "^4.18.2",
+    "jsonwebtoken": "^9.0.3"
```
fabiorodriguesdev: you forgot a comma here, that's why your build fails.


```suggestion
    "jsonwebtoken": "^9.0.3",
```

fabiorodriguesdev (changes requested):

DariaAleshina:
Hi! Impressive work💥! 

The only thing i could suggest is to check out with smw from backend-expert teachers (e.g. @FrontierPsychiatrist ) the error handing flow. I have just worked on err handling middleware and based on feedback removed try/catch blocks from userController completely to throw only generic 500 error to the client at that point. 

Also i don't think that exposing message 'JWT secret is not configured' to the client is a good idea (as it reveals internal configuration details) - should change to smth generic like "Smth went wrong". Btw i merged err handing middleware yesterday so you could update your branch with new main and use it for that.

backend/package.json
MaksVintskovsky: thanks! Fixed

fabiorodriguesdev (approved):

PR #78, opened 2026-04-24
42 jwt middleware for protected routes

freeediver:
- resolves #42

### DIFF
--- a/backend/src/config/env.ts
+++ b/backend/src/config/env.ts
@@ -1,0 +1,8 @@
+import 'dotenv/config';
+
+// Validate critical env vars BEFORE starting server
+export const JWT_SECRET = process.env.JWT_SECRET;
+if (!JWT_SECRET) {
+  console.error('❌ FATAL: JWT_SECRET is not set. Check your .env file.');
+  process.exit(1);
+}
--- a/backend/src/index.ts
+++ b/backend/src/index.ts
@@ -1,7 +1,9 @@
+import 'dotenv/config';
 import express, { Request, Response } from 'express';
 import helmet from 'helmet';
 import cors from 'cors';
 import userRouter from './routes/userRoutes.js';
+import './config/env.js';
 
 const app = express();
 
--- a/backend/src/middleware/authenticate.ts
+++ b/backend/src/middleware/authenticate.ts
@@ -1,4 +1,3 @@
-import 'dotenv/config';
 import { Request, Response, NextFunction } from 'express';
 import { JWT_SECRET } from '../config/env.js';
 import jwt from 'jsonwebtoken';
--- a/backend/src/middleware/authenticate.ts
+++ b/backend/src/middleware/authenticate.ts
@@ -1,5 +1,6 @@
 import 'dotenv/config';
 import { Request, Response, NextFunction } from 'express';
+import { JWT_SECRET } from '../config/env.js';
 import jwt from 'jsonwebtoken';
 
 export function authenticate(req: Request, res: Response, next: NextFunction) {
@@ -14,18 +15,11 @@ export function authenticate(req: Request, res: Response, next: NextFunction) {
   // 3. Extract the token
   const token = authHeader.substring(7);
 
-  // 4. Validate secret exists
   try {
-    const secret = process.env.JWT_SECRET;
-    if (!secret) {
-      console.error('FATAL: JWT_SECRET environment variable not set');
-      return res.status(500).json({ error: 'Server configuration error' });
-    }
+    // 4. Verify token
+    const decoded = jwt.verify(token, JWT_SECRET!);
 
-    // 5. Verify token
-    const decoded = jwt.verify(token, secret);
-
-    // 6. Validate payload structure
+    // 5. Validate payload structure
     if (
       typeof decoded !== 'object' ||
       decoded === null ||
@@ -36,7 +30,7 @@ export function authenticate(req: Request, res: Response, next: NextFunction) {
       return res.status(401).json({
         error: 'Invalid token format',
       });
-    } 
+    }
 
     // 7. Attach the payload to req.user and call next()
     req.user = decoded as { userId: string; role: string }; // Attach user data to request
--- a/backend/src/middleware/authenticate.ts
+++ b/backend/src/middleware/authenticate.ts
@@ -26,7 +26,7 @@ export function authenticate(req: Request, res: Response, next: NextFunction) {
     const decoded = jwt.verify(token, secret);
 
     // 6. Validate payload structure
-    /* if (
+    if (
       typeof decoded !== 'object' ||
       decoded === null ||
       !('userId' in decoded) ||
@@ -36,7 +36,7 @@ export function authenticate(req: Request, res: Response, next: NextFunction) {
       return res.status(401).json({
         error: 'Invalid token format',
       });
-    } */
+    } 
 
     // 7. Attach the payload to req.user and call next()
     req.user = decoded as { userId: string; role: string }; // Attach user data to request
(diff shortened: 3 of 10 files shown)


backend/src/middleware/authenticate.ts
```diff
@@ -0,0 +1,55 @@
+import 'dotenv/config';
+import { Request, Response, NextFunction } from 'express';
+import jwt from 'jsonwebtoken';
+
+export function authenticate(req: Request, res: Response, next: NextFunction) {
+  // 1. Read the Authorization header
```
ssda-developer: Maybe we should remove the unnecessary step-by-step comments?

backend/src/middleware/authenticate.ts
```diff
@@ -0,0 +1,55 @@
...
+      typeof decoded !== 'object' ||
+      decoded === null ||
+      !('userId' in decoded) ||
+      !('role' in decoded)
+    ) {
+      console.warn('Invalid token payload structure');
+      return res.status(401).json({
+        error: 'Invalid token format',
+      });
+    } */
```
ssda-developer: If this commented-out block is no longer needed, it might be better to remove it to keep the code clean.

backend/src/middleware/authenticate.ts
```diff
@@ -0,0 +1,55 @@
...
+    return res.status(401).json({ error: 'No token provided. Please log in.' });
+  }
+
+  // 3. Extract the token
+  const token = authHeader.substring(7);
+
+  // 4. Validate secret exists
+  try {
+    const secret = process.env.JWT_SECRET;
+    if (!secret) {
```
ssda-developer: Just an idea: maybe we should validate JWT_SECRET globally at server startup. If the secret is missing, the app shouldn't even start.

backend/src/middleware/authenticate.ts
freeediver: I just uncommented this bloc which is essential for validating the payload.👍

backend/src/middleware/authenticate.ts
freeediver: I think the step-by-step comments add useful clarity for both other developers and my future self revisiting the code months later. I’d prefer to keep them for readability.

backend/src/middleware/authenticate.ts
freeediver: Nice catch! I've just updated it.

backend/src/middleware/authenticate.ts
FrontierPsychiatrist: I agree with @ssda-developer. Comments should be added when something is not clear from the code, but this is not the case here.

For example below, where there is the `.substring(7)`, a comment could explain why "7". But this assignment doesn't really need a comment.

Main issue with too many comments is that they get out of sync with the code _fast_. People don't update them, it's a sad fact.

What could be added though is a top-level JSDoc comment explaing what this middelware even does!

backend/src/test-auth.ts
```diff
@@ -0,0 +1,57 @@
...
+
+// Add this to test expired token scenario: returns 401 with expired token
+app.get('/test-expired', authenticate, (req, res) => {
+  res.json({ message: 'This should not work with expired token' });
+});
+
+app.listen(4000, () => {
+  console.log('✅ Server running on http://localhost:4000');
+  console.log('   GET /public');
+  console.log('   GET /protected (use token above)');
```
FrontierPsychiatrist: What does this mean, "above"?

backend/src/config/env.ts
```diff
@@ -0,0 +1,8 @@
+import 'dotenv/config';
+
+// Validate critical env vars BEFORE starting server
+export const JWT_SECRET = process.env.JWT_SECRET;
+if (!JWT_SECRET) {
+  console.error('❌ FATAL: JWT_SECRET is not set. Check your .env file.');
```
FrontierPsychiatrist, line 6: Note, in deployed environments this doesn't necessarily come from a `.env` file, it can also be set in other ways.

backend/src/middleware/authenticate.ts
```diff
@@ -0,0 +1,49 @@
+import 'dotenv/config';
```
FrontierPsychiatrist: Unnecessary import

backend/src/routes/userRoutes.ts
```diff
@@ -1,13 +1,14 @@
...
 
 const userController = new UserController();
 const userRouter = Router();
 
 userRouter.get('/', (req, res) => userController.getUsers(req, res));
 userRouter.get('/:id', (req, res) => userController.getUserById(req, res));
-userRouter.post('/', (req, res) => userController.createUser(req, res));
-userRouter.put('/:id', (req, res) => userController.updateUser(req, res));
-userRouter.delete('/:id', (req, res) => userController.deleteUser(req, res));
+userRouter.post('/', authenticate, (req, res) => userController.createUser(req, res));
```
FrontierPsychiatrist: Hm, is this really correct? Does this work with new user onboarding?

backend/src/index.ts
```diff
@@ -1,7 +1,9 @@
+import 'dotenv/config';
```
FrontierPsychiatrist, line 1: Unnecessary import

backend/src/test-auth.ts
freeediver: @FrontierPsychiatrist Since we’re still in the development phase, I added test-auth.ts to test and debug the middleware and make sure everything is working correctly. This setup is only temporary, so we can remove it at any time.
To answer your question, app.listen(4000, ...) is what starts the server and allows it to listen for incoming HTTP requests defined in test-auth.ts on port 4000.

backend/src/routes/userRoutes.ts
freeediver: My mistake, i’ll remove this from the route.
If user creation is intended to be restricted to admins, then keeping this route with authentication would make sense. However, since the ticket doesn’t specify that requirement, I’ll remove it for now.

backend/src/middleware/authenticate.ts
freeediver: @FrontierPsychiatrist  Since I'm storing the JWT_SECRET in the .env file, i  need to import it to load the environment variables into process.env. Therefore, I'd like to know where environment variables should be managed and stored for our app. I'll ask about this during today's session.

backend/src/test-auth.ts
FrontierPsychiatrist: What does "use token above" mean? Where above is that token?

backend/src/middleware/authenticate.ts
FrontierPsychiatrist: You're already importing `../config/env.js` which is the module that's taking care of setting up `dotenv`, so it's not necessary again

backend/src/routes/userRoutes.ts
FrontierPsychiatrist: I'm not sure myself though. When is the endpoint called? During the signup process?

backend/src/middleware/authenticate.ts
freeediver: I already changed it. I discussed with Khaled today which strategy or tool we should use instead of .env, and he suggested using Docker to save and manage the environment variables. Any suggestions you have would be greatly appreciated.

backend/src/index.ts
freeediver: By removing this import, the environment variables stored in .env will not be taken into consideration. Including this import helps the index parse the .env file.

backend/src/test-auth.ts
freeediver: "Use token above" refers to the testing token. The first time, I tried to add a token as an example, but the JWT_SECRET should also be shared for testing. Therefore, I'll remove this line.  
Everyone can test with their own JWT_SECRET and token by following the instructions in the Backend README file.

backend/src/routes/userRoutes.ts
freeediver: I've already removed "authenticate" from this route. I think it's intended for the sign-up process.

fabiorodriguesdev (approved):

PR #80, opened 2026-04-29
61 Adding accessibility and ui improvements to the button component.

### LINKED ISSUE #61: Accessibility improvements for the button component
This ticket is work on the accessibility improvements that @DariaAleshina proposed : )

Hi! Great job, guys. There are quite a lot of variants to handle here 👏

I noticed a few things that might be worth reviewing or improving:

* **Keyboard accessibility in Storybook:** the button currently isn’t reachable via keyboard navigation (Tab). I understand that the `loading` and `disabled` states shouldn’t be focusable, but the default state should be. I wasn’t able to identify why this is happening — it might be a Storybook-specific issue — but it would be good to double-check.

* **Cursor style:** the default state seems to be missing `cursor: pointer`. Adding it would align with common UX expectations for clickable elements.

* **Hover transition:** it might be nice to add a transition duration (around **300ms**) to the hover animation. This is a common pattern and usually makes the interaction feel smoother.

* **Small variant text wrapping:** in the Storybook example for the `small` variant, the text wraps to the next line when it doesn’t fit, which doesn’t look great visually. It might be better to either:

  * allow the button to grow based on its content width, or
  * truncate the text instead (Tailwind has a `truncate` utility that could help with this).

_Originally posted by @DariaAleshina in https://github.com/ReDI-School/S26-Full-Stack-Circle/issues/36#issuecomment-4222666307_

santi-b-b:
Implements accessibility and UI improvements for the button component:

- Fixed keyboard navigation so enabled buttons are reachable via Tab
- Added cursor: pointer for interactive button states
- Added hover transition duration for smoother interaction feedback
- Prevented text wrapping in small button variant by truncating overflow text
---
- Closes #61

### DIFF
--- a/frontend/src/components/Button/Button.styles.ts
+++ b/frontend/src/components/Button/Button.styles.ts
@@ -7,10 +7,12 @@ export const buttonStyles = tv({
     'font-semibold',
     'font-sans',
     'text-white',
-    'focus-visible:outline-2',
-    'focus-visible:outline-offset-2',
-    'focus-visible:outline-white',
+    'focus-visible',
     'transition-all',
+    'cursor-pointer',
+    'transition-all duration-300',
+    'truncate',
+    'p-2.5',
   ],
   variants: {
     variant: {
--- a/frontend/src/components/Button/Button.styles.ts
+++ b/frontend/src/components/Button/Button.styles.ts
@@ -9,7 +9,6 @@ export const buttonStyles = tv({
     'text-white',
     'focus-visible',
     'transition-all',
-    'cursor-pointer',
     'transition-all duration-300',
     'whitespace-nowrap',
   ],
@@ -26,9 +25,9 @@ export const buttonStyles = tv({
       default: 'min-w-50 text-base py-3 px-6',
     },
     state: {
-      default: '',
-      disabled: 'cursor-not-allowed disabled:pointer-events-none disabled:opacity-60',
-      loading: 'opacity-70 cursor-wait pointer-events-none',
+      default: 'cursor-pointer',
+      disabled: 'cursor-not-allowed disabled:opacity-60 cursor-not-allowed',
+      loading: 'opacity-70 cursor-progress',
     },
   },
   compoundVariants: [
--- a/frontend/src/components/Button/Button.styles.ts
+++ b/frontend/src/components/Button/Button.styles.ts
@@ -26,7 +26,7 @@ export const buttonStyles = tv({
     },
     state: {
       default: 'cursor-pointer',
-      disabled: 'cursor-not-allowed disabled:opacity-60 cursor-not-allowed',
+      disabled: 'opacity-60 cursor-not-allowed',
       loading: 'opacity-70 cursor-progress',
     },
   },
--- a/frontend/src/components/Button/Button.styles.ts
+++ b/frontend/src/components/Button/Button.styles.ts
@@ -11,8 +11,7 @@ export const buttonStyles = tv({
     'transition-all',
     'cursor-pointer',
     'transition-all duration-300',
-    'truncate',
-    'p-2.5',
+    'whitespace-nowrap',
   ],
   variants: {
     variant: {
@@ -23,8 +22,8 @@ export const buttonStyles = tv({
       negative: 'bg-button-negative ',
     },
     size: {
-      small: 'w-[100px] h-[32px] text-sm',
-      default: 'w-[200px] h-[50px]',
+      small: 'min-w-25  text-sm py-2 px-4',
+      default: 'min-w-50 text-base py-3 px-6',
     },
     state: {
       default: '',


frontend/src/components/Button/Button.styles.ts
```diff
@@ -7,10 +7,12 @@ export const buttonStyles = tv({
     'font-semibold',
     'font-sans',
     'text-white',
-    'focus-visible:outline-2',
-    'focus-visible:outline-offset-2',
-    'focus-visible:outline-white',
+    'focus-visible',
     'transition-all',
+    'cursor-pointer',
```
Davies70: Nice accessibility improvement. Adding `cursor-pointer` makes the buttons feel more interactive. Glad to see this fixed from earlier.

frontend/src/components/Button/Button.styles.ts
```diff
@@ -7,10 +7,12 @@ export const buttonStyles = tv({
...
+    'transition-all duration-300',
+    'truncate',
+    'p-2.5',
```
fabiorodriguesdev: Why is this padding necessary? this adds extra padding making the positioning of the label wrong: 

Image

Buttons naturally has centered content, so if you want to add padding, you have to make sure you add different padding amounts for each size.

for the `truncate`, if a button has a sooo big label that needs to be truncated, the problem is not really the button but the content creator. Buttons basically never will have so big content that should truncate. to prevent making it break in two lines you can use `whitespace-nowrap`

frontend/src/components/Button/Button.styles.ts
santi-b-b: 😄

frontend/src/components/Button/Button.styles.ts
santi-b-b: Hey so I agree with the vertical padding I could kept it as only horizontal. My idea was to define a criteria until it was reasonable to have text in the button. Otherwise we could have text until the very limit which I also think is not the idea (check the first picture). I also prefer the truncate beacuse if we want to keep that horizontal padding it actually shows you how far you can go (check picture 2). Other wise you could get very close to the right margin of the button (check picture 3).

But this is just the way I think looks the best, what do you think? 

1:
image

2:
image

3:
image

santi-b-b: just to be extra clear, I prefer option 2

DariaAleshina: I would agree with Fabio. You are a bit overcomplicating it :) You use fixed width & height in px for the button, which causes limitations. 

You can approach button size differently by defining: 1) font size (+ line height) 2) padding, and ...3) min-width as we see per figma are 200px /100px. My humble suggestion would be to do smth like this:     
```
size: {
      small: 'min-w-25  text-sm py-2 px-4',
      default: 'min-w-50 text-base py-3 px-6',
```
This way you keep min width per design, and when text is too long button will grow in width accordingly and padding left/right will make sure there is always space between text and the border.

Also think of this edge case related to accessibility: font sizes use REM units, which are relative to the browser's root font size (16px by default). But users can increase this in their browser settings for accessibility reasons (e.g. visual impairment). If the button has a fixed height, it won't adapt,  the text may get clipped or overflow, because the height is set in pixels and doesn't respond to that change.

santi-b-b: I think rather than complicated or not, this is more of a design definition. we are all taking about different alternatives on how to handle a button that has a long text, but are we allowing that? 

if we always want the button to stay the same size, I still prefer the 2nd option I presented. If we want to have buttons with long texts I would agree to use a min-w and let it grow. 

about the accessibility and height I would also agree to let it grow to allow bigger font sizes.

DariaAleshina: tbh I just assumed that as Fabio confirmed that button should not get a too long text and truncate is not really needed this is the approach to refer to (and he is basically the project designer :). In any case my key point here is general approach to sizing the button, which e.g. in case of small variant currently causes content shift.

DariaAleshina (dismissed):

frontend/src/components/Button/Button.styles.ts
```diff
@@ -26,9 +25,9 @@ export const buttonStyles = tv({
       default: 'min-w-50 text-base py-3 px-6',
     },
     state: {
-      default: '',
-      disabled: 'cursor-not-allowed disabled:pointer-events-none disabled:opacity-60',
-      loading: 'opacity-70 cursor-wait pointer-events-none',
+      default: 'cursor-pointer',
+      disabled: 'cursor-not-allowed disabled:opacity-60 cursor-not-allowed',
```
DariaAleshina: now you have two `cursor-not-allowed` :). if you wanted to duplicate it in disabled state (which i am not really sure you need) you should put `disabled:` in front of it.

DariaAleshina (approved):

PR #81, opened 2026-04-30
feat: implement event retrieval by ID with corresponding routes and service

Davies70:
##  Add Get Single Event Endpoint

- Added `getEventById` service using Prisma with organizer relation
- Included organizer `firstName` and `lastName` in the response
- Implemented corresponding controller logic
- Created `GET /events/:id` route
- Wired up route in the server

- Closes #49


Davies70:
@FrontierPsychiatrist @fabi-rodrigues 

Hi both, apologies if I may have tagged the wrong reviewers.

I’ve pushed my latest commits for review and would appreciate any feedback when you get the chance.

Thanks for your time.

backend/src/controllers/eventController.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export class EventController {
+  getEventById = async (req: Request, res: Response) => {
+    const { id } = req.params;
+    const event = await eventService.getEventById(id);
+    if (!event) {
+      return res.status(404).json({ error: 'Event not found' });
+    }
+
+    res.json({ event });
```
ssda-developer, line 36: Hey! Good job.
Maybe we should wrap this in a try...catch to handle potential errors?

Davies70: Thank you for your input! This defeats the purpose of the error handling middleware. What do you think?

ssda-developer: Ah, I see! I was just looking at this specific snippet. If we have global error handling middleware in place, then try...catch would definitely be redundant.

santi-b-b: hey! but in this case do we need to use the next function in order for the middleware to get the error?

when testing the route https://redi-events-api-git-fe-50256a-fabio-rodrigues-projects-d0d7c49d.vercel.app/events/4 is giving me a generic 500 error in stead of the 404 you have there.

FrontierPsychiatrist: From a quick glance, no - I assume something else is going wrong here. @fabi-rodrigues would need to take a look in the backend logs.

Returning a response (the `return res.status(404)`) should not trigger the error middleware.

Since all ids are basically failing I assume there is some error in the eventService here that fails regardless of if  the event exists or not.

FrontierPsychiatrist: It is some issue with the database:

```
Error: PrismaClientKnownRequestError:
Invalid `prisma.event.findUnique()` invocation:


The table `public.Event` does not exist in the current database.
```

fabiorodriguesdev: I'm so confused who is @fabi-rodrigues LOL I'm not this person, I'm @fabiorodriguesdev !

Here's the log from vercel!

```bash
Error: PrismaClientKnownRequestError: 
Invalid `prisma.event.findUnique()` invocation:


The table `public.Event` does not exist in the current database.
    at Gr.handleRequestError (/var/task/backend/node_modules/@prisma/client/runtime/client.js:65:8286)
    at Gr.handleAndLogRequestError (/var/task/backend/node_modules/@prisma/client/runtime/client.js:65:7581)
    at Gr.request (/var/task/backend/node_modules/@prisma/client/runtime/client.js:65:7288)
    at process.processTicksAndRejections (node:internal/process/task_queues:104:5)
    at async a (/var/task/backend/node_modules/@prisma/client/runtime/client.js:75:6730)
    at async EventService.getEventById (file:///var/task/backend/dist/src/services/eventService.js:4:16)
    at async EventController.getEventById (file:///var/task/backend/dist/src/controllers/eventController.js:7:27) {
  code: 'P2021',
  meta: {
    modelName: 'Event',
    driverAdapterError: DriverAdapterError: TableDoesNotExist
        at PrismaPgAdapter.onError (file:///var/task/backend/node_modules/@prisma/adapter-pg/dist/index.mjs:651:11)
        at PrismaPgAdapter.performIO (file:///var/task/backend/node_modules/@prisma/adapter-pg/dist/index.mjs:646:12)
        at process.processTicksAndRejections (node:internal/process/task_queues:104:5)
        at async PrismaPgAdapter.queryRaw (file:///var/task/backend/node_modules/@prisma/adapter-pg/dist/index.mjs:566:30)
        at async e.interpretNode (/var/task/backend/node_modules/@prisma/client/runtime/client.js:11:44573)
        at async e.interpretNode (/var/task/backend/node_modules/@prisma/client/runtime/client.js:11:45017)
        at async e.interpretNode (/var/task/backend/node_modules/@prisma/client/runtime/client.js:11:43696)
        at async e.interpretNode (/var/task/backend/node_modules/@prisma/client/runtime/client.js:11:46237)
        at async e.run (/var/task/backend/node_modules/@prisma/client/runtime/client.js:11:43287)
        at async e.execute (/var/task/backend/node_modules/@prisma/client/runtime/client.js:57:815) {
      cause: [Object]
    }
  },
  clientVersion: '7.5.0'
}
```

fabiorodriguesdev: I discovered that when I use my company computer, my user is @fabi-rodrigues, annoying, will fix it later :)

fabiorodriguesdev: There you go!

https://redi-events-428mf2tcg-fabio-rodrigues-projects-d0d7c49d.vercel.app/events/2f7c3a83-f924-4ece-8107-0f62e606e47b

```json
{
  "error": "Event not found"
}
```

I had to add a `prisma db push` in the building phase to make it work with the table changes.

backend/src/controllers/eventController.ts
```diff
@@ -26,4 +26,17 @@ export class EventController {
...
+      const event = await eventService.getEventById(id);
+      if (!event) {
+        return res.status(404).json({ error: 'Event not found' });
+      }
+
+      res.json({ event });
+    } catch (err) {
+      next(err);
+    }
+  };
```
DariaAleshina: The try/catch isn't needed here — we have `express-async-errors` which automatically forwards unhandled rejections to error middleware. The `!event` 404 check is intentional business logic so it is not an Error that will be caught by error middleware. I believe it could be simplified to:
```
getEventById = async (req: Request, res: Response) => {
  const { id } = req.params;
  const event = await eventService.getEventById(id);
  if (!event) {
    return res.status(404).json({ error: 'Event not found' });
  }
  res.json({ event });
};
```

Davies70: This was my earlier implementation. After seeing the getEvents method merged into main with a try/catch, I updated getEventById to follow the same pattern, although it looks like this case didn’t actually need custom error mapping.

DariaAleshina (approved):
looks perfect to me! :)

PR #82, opened 2026-05-04
48 list all events get events

santi-b-b:
- Closes #48 

## Summary
Implemented optional date filtering for the `GET /events` endpoint.

Users can now filter events by whether they are upcoming or past using the `filter` query parameter.

## Changes Made
- Added `EventService.getEvents(filter?)` with Prisma date filtering
- Added support for:
  - `GET /events`
  - `GET /events?filter=upcoming`
  - `GET /events?filter=past`
- Added validation for invalid filter values
- Return `400 Bad Request` for unsupported filters
- Included organizer `firstName` and `lastName` in event responses
- Added event routes and registered them in `index.ts`

## Accepted Filters
- `upcoming`
- `past`

## Invalid Filter Behavior
Requests with invalid filter values return:

```json
{
  "error": "Invalid filter. Use \"upcoming\" or \"past\"."
}

### DIFF
--- a/backend/src/controllers/eventController.ts
+++ b/backend/src/controllers/eventController.ts
@@ -1,5 +1,6 @@
 import { Request, Response } from 'express';
 import { EventService } from 'src/services/eventService.js';
+import { NextFunction } from 'express';
 
 const eventService = new EventService();
 
@@ -13,23 +14,16 @@ function parseEventFilter(value: unknown): EventFilter | undefined {
 }
 
 export class EventController {
-  async getEvents(req: Request, res: Response) {
+  async getEvents(req: Request, res: Response, next: NextFunction) {
     try {
       const filter = parseEventFilter(req.query.filter);
       const events = await eventService.getEvents(filter);
       res.json({ events });
-    } catch (error) {
-      if (error instanceof Error && error.message === 'INVALID_EVENT_FILTER') {
-        return res.status(400).json({
-          error: 'Invalid filter. Use "upcoming" or "past".',
-        });
-      }
-
-      console.error('Error fetching events:', error);
-
-      res.status(500).json({
-        error: 'Failed to fetch events',
-      });
+    } catch (err) {
+      console.log('Error caught:', err);
+      console.log('Error type:', err instanceof Error);
+      console.log('Error message:', (err as Error).message);
+      next(err);
     }
   }
 }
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by santi-b-b** [comment]
@DariaAleshina hey! should I take the 500 error from this controller in the end? it is going to be handled in the middleware from what I saw

**Comment by DariaAleshina** [comment]
> @DariaAleshina hey! should I take the 500 error from this controller in the end? it is going to be handled in the middleware from what I saw

Hey! So my error handler middleware will **automatically** catch errors , and i guess if my error 500  with "smth went wrong" is fine for your case (which Moritz said if often pretty much enough for the client), then you can even remove try/catch in controller (as i did on userController). I would tho consult smw from backend team 😁 just in case

File: backend/src/middleware/errorHandler.ts [hunk 1]
```diff
@@ -15,5 +15,10 @@ export function errorHandler(
   next: NextFunction
 ) {
   console.error('Error:', err);
+
+  if (err.message === 'INVALID_EVENT_FILTER') {
+    return res.status(400).json({ error: 'Invalid event filter' });
+  }
+
```
santi-b-b, line 22: @FrontierPsychiatrist what do you think of passing this error to the middleware?

FrontierPsychiatrist: This is not an error as per definition of a middleware. I think it should stay here.

FrontierPsychiatrist: Sorry maybe got it I little wrong. It could be an error, but IMO it's too specific for the middleware. It's perfect in the specific controller where it applies (again, IMO, I know there's people who prefer more complex generic error middlewares)

backend/src/controllers/eventController.ts
```diff
@@ -0,0 +1,29 @@
...
+export class EventController {
+  async getEvents(req: Request, res: Response, next: NextFunction) {
+    try {
+      const filter = parseEventFilter(req.query.filter);
+      const events = await eventService.getEvents(filter);
+      res.json({ events });
+    } catch (err) {
+      console.log('Error caught:', err);
+      console.log('Error type:', err instanceof Error);
+      console.log('Error message:', (err as Error).message);
```
motiko, line 25:
```ts
if (err instanceof Error) {
  console.log('Error message:', err.message);
}
```

motiko (approved):
Error handling and logging could be managed better. But this will require a bigger refactoring effort it seems.
Overall could be merged as is IMO.

PR #83, opened 2026-05-07
feat: create Attendance model add relations

### LINKED ISSUE #53: Create the Attendance model
We need a way to track which users are attending which events. This is a join table — it connects a `User` to an `Event`.
 
### Implementation Notes
 
Add the `Attendance` model to `prisma/schema.prisma`:
 
```prisma
model Attendance {
  id        String   @id @default(uuid())
  userId    String
  eventId   String
  user      User     @relation(fields: [userId], references: [id])
  event     Event    @relation(fields: [eventId], references: [id])
  createdAt DateTime @default(now())
 
  @@unique([userId, eventId])
}
```
 
You also need to add the relations on both `User` and `Event`:
 
```prisma
model User {
  // ... existing fields
  attendances Attendance[]
}
 
model Event {
  // ... existing fields
  attendances Attendance[]
}
```
 
Then sync and regenerate:
 
```bash
npm run prisma:push
npm run prisma:generate
```
 
> The `@@unique([userId, eventId])` constraint means a user can only register once for the same event — the database will reject duplicates automatically.
 
### Acceptance Criteria
 
- [ ] `Attendance` model exists with `userId` and `eventId` foreign keys
- [ ] A user cannot be registered to the same event twice
- [ ] `npm run prisma:studio` shows the new `Attendance` table

Aarbei:
- resolves #53 

Implemented an Attendance model in the Prisma schema.
Changes made:
- created Attendance model in Prisma schema;
- added relations:
  - User to many Attendance records;
  - Event to many Attendance records;
- added unique constraint on [userId, eventId] to prevent duplicate attendance records;
- validated and synced schema by running npx prisma validate, npm run prisma:push, and npm run prisma:generate.

### DIFF
--- a/backend/prisma/schema.prisma
+++ b/backend/prisma/schema.prisma
@@ -20,6 +20,7 @@ model User {
   createdAt    DateTime @default(now())
   updatedAt    DateTime @updatedAt
   events       Event[]
+  attendances Attendance[]
 }
 
 model Event {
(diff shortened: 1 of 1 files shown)


fabiorodriguesdev (approved):
Good job!

PR #85, opened 2026-05-13
Feature update own profile

MaksimTrofimov:
Closes #45
This PR implements the user profile update feature with proper authorization control.

## Implemented changes
- Users can update only their own profile (firstName, lastName)
- Returns 403 Forbidden when attempting to update another user's profile
- Password hash is excluded from API response

## Notes on task scope
Some parts of the original task were already implemented in the codebase:

- 401 Unauthorized handling (no token provided) is already implemented in `backend/src/middleware/authenticate.ts`
- Route is already wired with `authenticate` middleware
- The `updateUser` method in both controller and service already existed before this task

The feature was tested using Postman.

### DIFF
--- a/backend/src/controllers/userController.ts
+++ b/backend/src/controllers/userController.ts
@@ -29,6 +29,11 @@ export class UserController {
 
   async updateUser(req: Request, res: Response) {
     const { id } = req.params;
+
+    if (!req.user || req.user.userId !== id) {
+      return res.status(403).json({ error: 'Forbidden' });
+    }
+
     const body = req.body;
     const user = await userService.updateUser(id, body);
 
--- a/backend/src/dto/user.dto.ts
+++ b/backend/src/dto/user.dto.ts
@@ -1,0 +1,15 @@
+import { User } from '@prisma/client';
+
+export class UserDTO {
+  id: string;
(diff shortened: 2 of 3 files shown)


backend/src/controllers/userController.ts
```diff
@@ -29,6 +29,11 @@ export class UserController {
 
   async updateUser(req: Request, res: Response) {
     const { id } = req.params;
+
+    if (!req.user || req.user.userId !== id) {
+      return res.status(403).json({ error: 'Forbidden' });
+    }
+
```
santi-b-b, line 36: hey Maskym, everything looks great! I only have a doubt about this. what happens if someone sends a request from outside the app, maybe with postman, I think with this validation someone could manually create a request and update a user, but I may be mistaken. 
 Maybe the user id has to come from the jwt in order to actually know the user is autenticated.

santi-b-b: I already saw that it is protected in the routes 

userRouter.put('/:id', authenticate, (req, res) => userController.updateUser(req, res));

everything looks great, good job : )

santi-b-b (approved):
looks great! : ) good job

PR #86, opened 2026-05-14
feat: add DELETE /users/:id endpoint

### LINKED ISSUE #46: Delete own account (`DELETE /users/:id`)
A logged-in user should be able to delete their own account. Same ownership check as the update endpoint.
 
### Implementation Notes
 
Add a `deleteUser` method to `src/services/userService.ts`:
 
```ts
async deleteUser(id: string) {
  // Use prisma.user.delete()
  /** your logic goes here **/
}
```
 
Add the controller method — remember to check ownership before deleting:
 
```ts
async deleteUser(req: Request, res: Response) {
  // 1. Check req.user.userId matches the id param — if not, return 403
  // 2. Call the service
  // 3. Return 204 No Content on success
  /** your logic goes here **/
}
```
 
Wire up the route with the `authenticate` middleware:
 
```ts
userRouter.delete('/:id', authenticate, (req, res) => /** your logic goes here **/);
```
 
> `204 No Content` is the standard response for a successful delete — it means "it worked, and there's nothing to return".
 
### Acceptance Criteria
 
- [ ] `DELETE /users/:id` deletes the user's account
- [ ] Returns `403` if the logged-in user tries to delete someone else's account
- [ ] Returns `401` if no token is provided
- [ ] Returns `204` on success

ssda-developer:
Close #46

### DIFF
--- a/backend/src/controllers/userController.ts
+++ b/backend/src/controllers/userController.ts
@@ -37,8 +37,17 @@ export class UserController {
 
   async deleteUser(req: Request, res: Response) {
     const { id } = req.params;
-    await userService.deleteUser(id);
+    const authenticatedUser = req.user;
 
-    res.json({ message: 'User deleted successfully' });
+    if (!authenticatedUser) {
+      return res.status(401).json({ error: 'Unauthorized' });
+    }
+
+    if (authenticatedUser.userId !== id) {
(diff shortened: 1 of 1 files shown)


santi-b-b (approved):
Looks good! : )

Antonov548 (approved):

PR #87, opened 2026-05-14
Feat/list attendees event

### LINKED ISSUE #56: List attendees of an event (`GET /events/:id/attendees`)
Returns the list of users registered for a given event. This can be a public endpoint — no login required.
 
### Implementation Notes
 
Add a `getAttendees` method to `src/services/attendanceService.ts`:
 
```ts
async getAttendees(eventId: string) {
  // Use prisma.attendance.findMany() with eventId as filter
  // Include the user's firstName and lastName (exclude passwordHash)
  /** your logic goes here **/
}
```
 
Add the controller method:
 
```ts
async getAttendees(req: Request, res: Response) {
  // Get the eventId from req.params
  // Return 404 if the event does not exist
  /** your logic goes here **/
}
```
 
Add the route to `src/routes/eventRoutes.ts`:
 
```ts
eventRouter.get('/:id/attendees', (req, res) => /** your logic goes here **/);
```
 
### Acceptance Criteria
 
- [x] `GET /events/:id/attendees` returns the list of attendees
- [x] Each attendee shows `firstName` and `lastName` but not `passwordHash`
- [x] Returns `404` if the event does not exist
- [x] No authentication required

MaksimTrofimov:
closed #56


backend/src/controllers/eventController.ts
```diff
@@ -26,4 +27,17 @@ export class EventController {
...
+    const event = await eventService.getEventById(eventId);
+
+    if (!event) {
+      return res.status(404).json({ error: 'Event does not exist' });
+    }
+
+    const attendees = await attendanceService.getAttendees(eventId);
+
+    res.json({ attendees });
+  }
```
santi-b-b, line 57: Hey, one doubt why is there an attendanceService file and not and attendanceController one? I would have created a controller and route too if necessary and if not I would kept all the attendance logic on the event structure, including the service

MaksimTrofimov: > Hey, one doubt why is there an attendanceService file and not and attendanceController one? I would have created a controller and route too if necessary and if not I would kept all the attendance logic on the event structure, including the service

I followed the task requirements: it explicitly mentioned adding the route (`eventRouter.get('/:id/attendees', (req, res) => /** your logic goes here **/)`), so I continued using the existing `eventController` instead of creating a separate `attendanceController`.

My understanding was that attendance is part of the event logic, since users register specifically for an event. Attendance cannot exist independently without an event, so I considered this functionality to logically belong to the `eventController` and the `event` structure. For the same reason, I did not see a need to create a separate `attendanceController`, since attendance is not an independent part of the system that is used separately from events.

At the same time, the task also specified using `attendanceService`, which is probably why the question came up about having two services but only one controller.

Most likely, the Single Responsibility Principle  was applied here. `EventService` is responsible for event-related operations, while `AttendanceService` handles attendance-related logic. This kind of separation makes the architecture cleaner, simplifies maintenance, and helps isolate business logic by domain responsibility.

However, personally, I think both approaches would be valid in this case:

- using a single `EventService`;
- or splitting the logic between `EventService` and `AttendanceService`.

At the moment, the attendance logic is relatively small and tightly coupled to events, so using a single service would also be perfectly reasonable. However, the project is not finished yet, and attendance may contain more business logic in the future, so the services may have been separated in advance for scalability and maintainability reasons.

In any case, in my opinion, using a single controller was the correct decision, while the services themselves could reasonably be organized either as one shared service or as two separate services.

santi-b-b: Ok I understand, personally I would have had either 2 controllers and 2 services or 1 and 1, but it also looks good this way.

santi-b-b (dismissed):
good work : )

backend/src/controllers/eventController.ts
```diff
@@ -1,8 +1,10 @@
 import { Request, Response } from 'express';
 import { EventService } from 'src/services/eventService.js';
+import { NextFunction } from 'express';
```
santi-b-b: @MaksimTrofimov the backend check is failing because of this line, could you remove it and check again?

MaksimTrofimov: > @MaksimTrofimov the backend check is failing because of this line, could you remove it and check again?

Hi!

Actually, this NextFunction import was introduced by you about 2 weeks ago (on May 8th, 2026, at 10:47 AM), as clearly shown in the file history.

It looks like this line was brought back due to an incorrect conflict resolution during your recent branch merge, which is what caused the backend check to fail. Prior to this merge, when I pushed my feature branch, all GitHub CI checks (including the Vercel Backend  were completely green and passing successfully.

Since working with this function was originally your task, you are in the best position to know whether it is still needed or should be removed. Could you please fix this import on your side?

Additionally, please double-check your entire recent merge to ensure that other conflicts were resolved correctly and make any necessary adjustments, to make sure other parts of the code aren't broken.

backend/src/controllers/eventController.ts
MaksimTrofimov: I see your point, but the main reason I didn't split it into two controllers is to avoid over-engineering and follow the YAGNI (You Aren't Gonna Need It) principle.

In REST API design, nested resources like /events/:id/attendees are standardly managed by the parent controller (eventController) as long as the sub-resource doesn't have its own independent root context. Creating a separate controller and route file for just a single endpoint would generate unnecessary boilerplate code without adding any real value at this stage.

The controller's job is just to hand over the HTTP request to the service layer. Since the endpoint is physically a part of the eventRouter, keeping it in eventController keeps the routing layer clean, lean, and centralized. We should only split controllers when the attendance domain grows large enough and scales into a completely independent context.

https://medium.com/@raksbisht/understanding-the-yagni-principle-in-software-development-8ca19fcd39a9

santi-b-b (dismissed):
Looks good : )

DariaAleshina (approved):
looks great!

PR #89, opened 2026-05-18
delete an event 

wms198:
[Delete an event](https://github.com/ReDI-School/S26-Full-Stack-Circle/issues/52)

##  204
IMG_4845

## 403
IMG_8468

## 404
IMG_6055


backend/src/services/eventService.ts
```diff
@@ -1,5 +1,7 @@
 import prisma from 'src/libs/prisma.js';
+//import prisma from '../libs/prisma.js';
```
ssda-developer: For code cleanliness and maintainability, I think we should try to remove unnecessary comments.

backend/src/routes/eventRoutes.ts
```diff
@@ -1,9 +1,11 @@
 import { Router } from 'express';
-import { EventController } from 'src/controllers/eventController.js';
+import { EventController } from '../controllers/eventController.js';
+import { authenticate } from '../middleware/authenticate.js';
 
 const eventController = new EventController();
 const eventRouter = Router();
 
 eventRouter.get('/', (req, res, next) => eventController.getEvents(req, res, next));
+eventRouter.delete('/:id', authenticate, (req, res) => eventController.deleteEvent(req, res));
```
ssda-developer, line 13: I noticed we're missing the `next` function here. We should probably add it, because without it, if an exception occurs (for example, a database connection failure), Express won't catch it, and the request will likely hang.

wms198: No, I do not think so. I just shut down the docker compose and logged exactly what suppose to log from error handling.
image
If you have more question, you can check for this link: https://stackoverflow.com/questions/13133071/express-next-function-what-is-it-really-for

DariaAleshina (dismissed):
looks good to me :)

MaksimTrofimov (approved):
LGTM! Everything is green and clean.

PR #90, opened 2026-05-19
Feat/issue 50/create event endpoint

akhilachiju:
Resolves #50 

A logged-in user can create a new event. The organizer is set automatically from the token — the client does not send it.

What changed
- Added createEvent method to eventService.ts using prisma.event.create()
- Updated eventController.ts to read organizerId from req.user.userId instead of req.body
- Added authenticate middleware to POST /events in eventRoutes.ts

Why
To protect the create event route so only logged-in users can create events, and to ensure the organizer is always set from the token — not from client input.

Acceptance Criteria
- POST /events creates a new event
- organizerId is taken from the JWT token, not from the request body
- Returns 401 if no token is provided
- Returns 201 with the created event


backend/src/controllers/eventController.ts
```diff
@@ -29,16 +29,46 @@ function parseEventFilter(value: unknown): {
...
+        description,
+        date: new Date(date),
+        location,
+        capacity,
       });
+
+      res.status(201).json({ event });
+    } catch (error) {
+      console.error('Error creating event:', error);
+      res.status(500).json({ error: 'Failed to create event' });
```
DariaAleshina: hi! we have an error handling middleware now that will catch errors automatically and throw status 500 with a consistent message. you can (and should) remove try/catch block here completely (as catch block not needed - you repeat the middleware logic basically)

backend/src/routes/eventRoutes.ts
```diff
@@ -7,8 +7,9 @@ const eventController = new EventController();
 const attendanceController = new AttendanceController();
 const eventRouter = Router();
 
+eventRouter.get('/', (req, res, next) => eventController.getEvents(req, res, next));
+eventRouter.post('/', authenticate, (req, res) => eventController.createEvent(req, res));
```
DariaAleshina: hey hey! please add a user/form input validation middleware here (i guess after `authenticate`). 
for this import:
```
import { createEventSchema } from '../validators/schemas.js';
import { validate } from 'src/middleware/validateUserInput.js';
```
and add it to your route similar to `login` logic: 
```
authRouter.post('/login', validate(loginSchema), (req, res, next) =>
);
```
if middleware not available, please update your branch with fresh main

backend/src/controllers/eventController.ts
```diff
@@ -29,16 +29,46 @@ function parseEventFilter(value: unknown): {
...
+          error: 'Invalid event filter',
+        });
+      }
+
+      const events = await eventService.getEvents(filter);
+      res.json({ events });
+    } catch (err) {
+      next(err);
+    }
+  }
```
DariaAleshina: no need in try/catch block here eithter

backend/src/controllers/eventController.ts
akhilachiju: Done! Removed the try/catch block completely — errors will now be caught and handled consistently by the error handling middleware.

backend/src/routes/eventRoutes.ts
akhilachiju: Done! Added validate(createEventSchema) after authenticate on POST /events, following the same pattern as the login route. Also imported validate and createEventSchema in the routes file.

backend/src/controllers/eventController.ts
akhilachiju: Done! Removed the try/catch from getEvents as well — all errors are now handled by the error middleware consistently.

akhilachiju:
> @akhilachiju hey! Great job! I just left you couple of comments, one related to "error handling" (as we have a middleware for it), and request to add "validate" middleware (as I could not add it myself to endpoints as yet not available and merged to main already). Thank you!

Thank you for the review! All done — removed try/catch blocks from both getEvents and createEvent, and added the validate(createEventSchema) middleware after authenticate on POST /events. Changes are pushed!

backend/src/controllers/eventController.ts
```diff
@@ -29,18 +29,39 @@ function parseEventFilter(value: unknown): {
 }
 
 export class EventController {
-  async getEvents(req: Request, res: Response) {
+  async getEvents(req: Request, res: Response, next: NextFunction) {
```
DariaAleshina: i guess next function is unused now which m ight also cause errors (same in create event)

akhilachiju: Good catch! Removed next: NextFunction from both getEvents and createEvent since it's no longer used after removing the try/catch blocks. Also cleaned up the import.

DariaAleshina (approved):
🔥

PR #91, opened 2026-05-19
Feat/register for event

akhilachiju:
Resolves #54 

A logged-in user can register themselves for an event. Registration is rejected if the event is already at full capacity.

What changed
- Added attendanceService.ts with an attend method that checks event existence, validates capacity, and creates the attendance record
- Added attendanceController.ts to read eventId from req.params and userId from the JWT token, returning 201 on success or 400 on failure
- Added POST /:id/attend route to eventRoutes.ts protected by authenticate middleware

Why
To allow logged-in users to register for events, while preventing over-registration when an event is full or a user tries to register twice.

Acceptance Criteria
- POST /events/:id/attend registers the logged-in user for the event
- Returns 400 if the event is already at full capacity
- Returns 400 if the user is already registered
- Returns 401 if no token is provided
- Returns 201 on success


backend/src/controllers/attendanceController.ts
```diff
@@ -0,0 +1,34 @@
...
+          res.status(404).json({ error: 'Event not found' });
+          return;
+        }
+        if (error.message === 'EVENT_FULL') {
+          res.status(400).json({ error: 'Event is already at full capacity' });
+          return;
+        }
+      }
+      // Prisma unique constraint violation = already registered
+      res.status(400).json({ error: 'You are already registered for this event' });
```
santi-b-b: Correct me if I'm wrong, but here you check first if the event doesn't exist then if the event is full and if the error is not any of does you're assuming that is because the user is already in the event attendance?

Is that right?

akhilachiju: You're right! I was relying on the Prisma unique constraint violation to catch duplicate registrations, which meant any unexpected error would return a misleading "already registered" message. Fixed it by adding an explicit 
check in the service before the insert, and the controller now handles ALREADY_REGISTERED explicitly with a proper 500 fallback for unexpected errors.

santi-b-b (approved):
Good work 💪

PR #92, opened 2026-05-19
51-Update an Event

MaksVintskovsky:
Closes #51 
It's just a first commit to create a PR. in prosses


backend/src/controllers/eventController.ts
```diff
@@ -54,4 +55,33 @@ export class EventController {
...
+      if (event.organizerId !== userId) {
+        return res.status(403).json({
+          error: 'You are not allowed to update this event',
+        });
+      }
+      const updatedEvent = await eventService.updateEvent(eventId, updateData);
+      return res.status(200).json(updatedEvent);
+    } catch (err) {
+      next(err);
+    }
```
santi-b-b: I don't think this try catch is necessary. I think the errors will go automatically to the middlewaer. 

correct me if I'm wrong @DariaAleshina

backend/src/routes/eventRoutes.ts
```diff
@@ -1,10 +1,13 @@
 import { Router } from 'express';
 import { EventController } from 'src/controllers/eventController.js';
-
+import { authenticate } from 'src/middleware/authenticate.js';
 const eventController = new EventController();
 const eventRouter = Router();
 
-eventRouter.get('/', (req, res) => eventController.getEvents(req, res));
+eventRouter.get('/', (req, res, next) => eventController.getEvents(req, res, next));
```
santi-b-b: I think this is the error you're getting the getEvents function does not receive a next property anymore

@MaksVintskovsky

MaksVintskovsky: Thanks for the support! understand the problem now. I did some mess with the merging, will check it from home later.

santi-b-b (dismissed):
Looks great, good job!

backend/src/routes/eventRoutes.ts
```diff
@@ -1,11 +1,15 @@
...
 
 const eventController = new EventController();
+const attendanceController = new AttendanceController();
 const eventRouter = Router();
 
+eventRouter.post('/:id/attend', authenticate, (req, res) => attendanceController.attend(req, res));
 eventRouter.get('/', (req, res) => eventController.getEvents(req, res));
-
 eventRouter.get('/:id', (req, res) => eventController.getEventById(req, res));
+eventRouter.put('/:id', authenticate, (req, res) => eventController.updateEvent(req, res));
```
DariaAleshina: hey hey! please add a user/form input validation middleware here (i guess after `authenticate`). 
for this import:
```
import { updateEventSchema } from '../validators/schemas.js';
import { validate } from 'src/middleware/validateUserInput.js';
```
and add it to your route similar to `login` logic: 
```
authRouter.post('/login', validate(loginSchema), (req, res, next) =>
);
```

DariaAleshina: if middleware not available, please update your branch with fresh main

backend/src/controllers/attendanceController.ts
```diff
@@ -0,0 +1,37 @@
...
+        if (error.message === 'EVENT_FULL') {
+          res.status(400).json({ error: 'Event is already at full capacity' });
+          return;
+        }
+        if (error.message === 'ALREADY_REGISTERED') {
+          res.status(400).json({ error: 'You are already registered for this event' });
+          return;
+        }
+      }
+      res.status(500).json({ error: 'Internal server error' });
```
DariaAleshina, line 34: you should catch such errors with the error handler middleware (as we have it for this purpose + consistency). for this use `next(error) `i think

MaksVintskovsky: hey! thanks for your catch. I think Akilla was working on this ticket. I did rebase here, because i did some sabotage while merging... :)

backend/src/routes/eventRoutes.ts
MaksVintskovsky: done. thanks for the review. check it again please

backend/src/services/eventService.ts
```diff
@@ -1,6 +1,14 @@
...
 import { Prisma } from 'generated/prisma/client.js';
 
+type UpdateEventData = {
+  title?: string;
+  description?: string | null;
+  date?: Date;
+  location?: string;
+  capacity?: number;
+};
+
```
MaksimTrofimov: This type should be moved into a separate interface inside the backend/src/types folder to maintain the project architecture and keep the current file clean.

backend/src/controllers/eventController.ts
```diff
@@ -73,6 +74,39 @@ export class EventController {
...
 
-    res.json({ event });
+    return res.json({ event });
   };
+
+  async updateEvent(req: Request, res: Response) {
+    const eventId = req.params.id;
+    const userId = req.user?.userId;
+
+    if (!userId) {
```
MaksimTrofimov: This if check is redundant. If the token or userId is missing, the request won't even reach this controller (see middleware/authenticate.ts, line 37).

backend/src/controllers/eventController.ts
```diff
@@ -73,6 +74,39 @@ export class EventController {
       });
     }
 
-    res.json({ event });
+    return res.json({ event });
   };
+
+  async updateEvent(req: Request, res: Response) {
+    const eventId = req.params.id;
+    const userId = req.user?.userId;
```
MaksimTrofimov: The optional chaining (?.) is not needed here because req.user and userId are guaranteed to exist. If they didn't, the request wouldn't even reach this controller (see middleware/authenticate.ts, line 37).

backend/src/controllers/eventController.ts
MaksVintskovsky: You're right that authenticate guarantees req.user at runtime. TypeScript still sees it as optional because Request.user is declared with ?, so I'll switch to req.user!.userId and remove the redundant authorization check.

MaksimTrofimov (approved):
Looks Good To Me

PR #93, opened 2026-05-19
feature: add test data to the tables users, events and attendances

Aarbei:
Closes #58

Was done:
- Added seed data arrays for users, events, and attendances;
- Linked users to events for some attendance records;
- Completed npm run prisma:seed without errors.


backend/prisma/schema.prisma
```diff
@@ -11,40 +11,45 @@ datasource db {
...
-  email        String   @unique
-  firstName    String
-  lastName     String
-  passwordHash String
-  role         Role     @default(USER)
-  createdAt    DateTime @default(now())
-  updatedAt    DateTime @updatedAt
+  id           String       @id @default(uuid())
+  email        String       @unique
+  firstName    String       @map("first_name")
```
DariaAleshina: hey hey! can you explain the  mapping you added in the schema (`@map ` ), it is not part of ticket, i wonder what it is really changing. i am just curious (not an expert here at all)

backend/prisma/seed.ts
```diff
@@ -11,31 +11,165 @@ const prisma = new PrismaClient({ adapter });
...
+    });
+
+    createdAttendances.push(createdAttendance);
+  }
 
-  console.log('Test user created:', testUser);
-  console.log('Login data:');
-  console.log('email: test@example.com');
-  console.log('password: 123456');
-  console.log('Seed completed successfully!');
```
DariaAleshina, line 86: i would return this log as it indicates that the whole script has finished executing (very conventional)

backend/prisma/schema.prisma
```diff
@@ -11,40 +11,45 @@ datasource db {
...
-  event Event @relation(fields: [eventId], references: [id])
+  id        String   @id @default(uuid())
+  userId    String   @map("user_id")
+  eventId   String   @map("event_id")
+  user      User     @relation(fields: [userId], references: [id])
+  event     Event    @relation(fields: [eventId], references: [id])
   createdAt DateTime @default(now()) @map("created_at")
-  @@map ("attendance")
+
+  @@map("attendances")
```
Davies70: Hello Aliona, great implementation overall. The schema and seed structure look very clean. I noticed the seed is failing on the `Attendance` table because of the `@@map("attendances")` mapping. Since Prisma already handles the model name correctly, the mapping may not be necessary here. Removing it (or aligning it with the existing database table name (`@@map("attendance")`) should fix the seeding issue.

backend/prisma/seed.ts
DariaAleshina: i mean this one `console.log('Seed completed successfully!');`

backend/prisma/schema.prisma
Aarbei: Hi! Just saw in a tutorial that snake_case is a standard for db layer and I decided to add mapping, but after reading comments I've removed the mapping entirely, you're right, it's not  the part of the ticket:)

backend/prisma/seed.ts
Aarbei: Thanks! I have returned this log

backend/prisma/schema.prisma
Aarbei: Hi! Thanks, I've removed the mapping, so now it should be fine:)

MaksimTrofimov (approved):
Looks good, thanks!

PR #94, opened 2026-05-19
feat: implement registration endpoint #40

Tatianatod:
closes #40 
Implemented POST /auth/register endpoint


backend/src/controllers/authController.ts
```diff
@@ -24,4 +24,27 @@ export class AuthController {
...
+
+  async register(req: Request, res: Response, next: NextFunction) {
+    try {
+      const { email, firstName, lastName, password } = req.body;
+
+      if (!email || !firstName || !lastName || !password) {
+        return res
+          .status(400)
+          .json({ error: 'Email, first name, last name, and password are required' });
+      }
```
DariaAleshina: hi! you can skip the check here as user input will be checked and processed by the validation middleware i am currently working on (smw will just need to add it later on when both yours and mine PR will be merged to main).

backend/src/controllers/authController.ts
```diff
@@ -24,4 +24,27 @@ export class AuthController {
...
+
+      const newUser = await this.authService.register(email, firstName, lastName, password);
+      return res.status(201).json(newUser);
```
DariaAleshina: as per ticket requirements,  password should not be returnedto the client, and newUser object here contains it as per line above

backend/src/services/authService.ts
```diff
@@ -33,4 +32,29 @@ export class AuthService {
...
+    const newUser = await prisma.user.create({
+      data: {
+        email,
+        firstName,
+        lastName,
+        passwordHash,
+      },
+    });
+
+    const { passwordHash: _, ...userWithoutPassword } = newUser;
```
DariaAleshina: looks over-engineered 😅 taken you will not need passwordHash anywhere anymore

backend/src/controllers/authController.ts
Tatianatod: Hi Daria. Thanks. That's exactly what I did.

backend/src/controllers/authController.ts
Tatianatod: Yes, that is important. Thank you.

backend/src/controllers/authController.ts
```diff
@@ -24,4 +24,26 @@ export class AuthController {
...
     }
   }
+
+  async register(req: Request, res: Response, next: NextFunction) {
+    try {
+      const { email, firstName, lastName, password } = req.body;
+
+      const newUser = await this.authService.register(email, firstName, lastName, password);
+
+      const userResponse = { ...newUser } as any;
```
DariaAleshina, line 34: hey hey  :) i know i said that prev version looked like over-engeneering but now (taken this type script error) i think the direction was perfectly fine and as simple as it could get in plain JS:
`const { passwordHash, password, ...userResponse } = newUser;`

backend/src/routes/authRoutes.ts
```diff
@@ -10,4 +10,6 @@ authRouter.post('/login', validate(loginSchema), (req, res, next) =>
   authController.login(req, res, next)
 );
 
+authRouter.post('/register', (req, res, next) => authController.register(req, res, next));
+
```
DariaAleshina: As now you have a fresh main with my validation middleware, please add it - see the `login` as example above. You will need additionally to import **registerSchema** from _'../validators/schemas.js'_ and add `validate(registerSchema)` after '/register'. The middleware will check user input coming from fronend and throw relevant errors. Thank you! 🔥

backend/src/services/authService.ts
```diff
@@ -35,6 +31,28 @@ export class AuthService {
...
+  async register(email: string, firstName: string, lastName: string, password: string) {
+    const existingUser = await prisma.user.findUnique({
+      where: { email },
+    });
+
+    if (existingUser) {
+      throw new Error('EMAIL_ALREADY_IN_USE');
+    }
+
+    const saltRounds = 10;
```
DariaAleshina, line 45: i was wondering, what is this variable with some magical number? why is it called saltRounds? i am just curious 😅 never dealt with bcrypt

backend/src/services/authService.ts
```diff
@@ -2,12 +2,8 @@ import prisma from '../libs/prisma.js';
 import bcrypt from 'bcrypt';
 import jwt from 'jsonwebtoken';
 
-interface LoginResponse {
```
DariaAleshina, line 5: just curious, why did you remove this type definition in login and now the return value is also not an object with a token property and just a token string, does it not interrupt the logic in login controller?

backend/src/controllers/authController.ts
```diff
@@ -24,4 +24,26 @@ export class AuthController {
...
+      delete userResponse.passwordHash;
+      delete userResponse.password;
```
MaksVintskovsky: I think newUser from Prisma does NOT contain password. so this line is unnessesary, or not?

backend/src/controllers/authController.ts
```diff
@@ -24,4 +24,26 @@ export class AuthController {
       next(error);
     }
   }
+
+  async register(req: Request, res: Response, next: NextFunction) {
+    try {
+      const { email, firstName, lastName, password } = req.body;
```
MaksimTrofimov, line 30: All incoming data from any external source should be validated before being used in the application.

**This means the API currently accepts invalid or unexpected data such as:**

- invalid email format
- empty password
- wrong data types 
- missing required fields
- overly long values

The project already has **Zod** installed, so it could be used for validating incoming data before passing it to the service layer.

Please note that validation is already used during login.

backend/src/services/authService.ts
```diff
@@ -2,12 +2,8 @@ import prisma from '../libs/prisma.js';
...
-  token: string;
-}
-
 export class AuthService {
-  async login(email: string, password: string): Promise<LoginResponse> {
+  async login(email: string, password: string) {
     const user = await prisma.user.findUnique({
```
MaksimTrofimov: This code is duplicated because it is used both during registration and login:

const user = await prisma.user.findUnique({
  where: { email },
});

It would be better to extract it into a separate method, for example findUserByEmail, to avoid code duplication and follow the DRY (Don't Repeat Yourself) principle.

This would improve readability, simplify maintenance, and make future changes easier.

**I have already added the findUserByEmail method in the UserService**

backend/src/controllers/authController.ts
MaksimTrofimov: delete userResponse.password → redundant line (the field does not exist in the backend/prisma/schema.prisma model)

backend/src/controllers/authController.ts
DariaAleshina: hey i just merged main and added the input validation middleware to the router, so this is covered now

MaksimTrofimov (approved):
Looks good to me!

backend/src/controllers/authController.ts
Tatianatod: Hi Daria, yes, you're right.

backend/src/services/authService.ts
Tatianatod: I hadn't heard of it either before our project :) You can read about it, for example, [here](https://medium.com/@nikhilnainwaya17/understanding-how-bcrypt-algorithm-works-2e4082ad254f)

backend/src/services/authService.ts
Tatianatod: I updated the controller to work with the string directly. But you're right, keeping the explicit interface is much cleaner. Thanks!

backend/src/controllers/authController.ts
Tatianatod: You're right, password isn't in the database model. This line was redundant, thanks.

backend/src/controllers/authController.ts
Tatianatod: Thanks Maksim, Daria! I actually had some issues with Zod on my local setup earlier, but managed to fix everything.

backend/src/services/authService.ts
Tatianatod: Yes, you're right. Thank you for the help.

PR #106, opened 2026-05-20
18 create protectedlayout

### LINKED ISSUE #18: Create ProtectedLayout
### Description

Image

- Desktop: https://www.figma.com/design/itAzcC43MCVpSyJCcmQJgu/ReDi-Events?node-id=0-1
- Mobile: https://www.figma.com/design/itAzcC43MCVpSyJCcmQJgu/ReDi-Events?node-id=1-869

Create a `ProtectedLayout` component used as the shared layout for all authenticated pages. The layout consists of:

- **Header bar:** Logo (`REDI. EVENTS.`) on the left, `UserArea` component (avatar + name + dropdown) on the right
- **Content area:** Scrollable area below the header where page content (`children`) is rendered

This layout only provides the header. All page-specific content (tabs, buttons, grids, forms, etc.) is the responsibility of the child page.

### Props

```typescript
interface ProtectedLayoutProps {
  children: React.ReactNode;
}
```

### Acceptance Criteria

- [ ] **Responsive:**
  - [ ] Desktop: `Logo` displays full version ("REDI. EVENTS.")
  - [ ] Mobile: `Logo` displays compact version ("R. E."), `UserArea` shows avatar + chevron only (no name text)

freeediver:
- resolves #18


frontend/src/components/ProtectedLayout/ProtectedLayout.tsx
```diff
@@ -0,0 +1,37 @@
...
+import { Skeleton } from '@components/Skeleton';
+
+export default function ProtectedLayout({ children }: ProtectedLayoutProps) {
+  const { user, loadingAuth, signOut, goToProfile } = useAuth();
+  const { main, headerBar, content } = protectedLayoutStyles();
+
+  return (
+    <main className={main()}>
+      <header className={headerBar()}>
+        <Logo />
```
DariaAleshina: Hi! I guess you forgot to make Logo change to compact variant on mobile:
Image

Figma design:

Image

frontend/src/components/ProtectedLayout/ProtectedLayout.tsx
```diff
@@ -0,0 +1,37 @@
...
+        {loadingAuth ? (
+          <div className="flex items-center gap-3">
+            <Skeleton width={40} height={40} radius="full" />
+            <div className="hidden md:block">
+              <Skeleton width={100} height={16} radius="base" />
```
DariaAleshina, line 26: ah nice addition (it was not in the ticket). could maybe attach a screenshot to the PR to show how it will look like in loading state? or add a story (if its possible)

frontend/src/components/ProtectedLayout/ProtectedLayout.styles.ts
```diff
@@ -0,0 +1,25 @@
...
+      'justify-between',
+      'items-center',
+      'px-7.5',
+      'lg:px-12.5',
+      'py-4',
+      'bg-white',
+      'border-b',
+      'border-gray-200',
+    ],
+    content: ['flex-1', 'overflow-y-auto'],
```
DariaAleshina: i would add  - at least horizontal - padding here as well (as in header bar) cause it will be consistent. 

Image

frontend/src/components/ProtectedLayout/ProtectedLayout.styles.ts
```diff
@@ -0,0 +1,25 @@
+import { tv } from 'tailwind-variants';
+
+export const protectedLayoutStyles = tv({
+  slots: {
+    main: ['flex', 'flex-col', 'h-screen', 'overflow-hidden'],
+    headerBar: [
+      'sticky',
+      'top-0',
+      'left-0',
+      'right-0',
```
DariaAleshina: why do you need "sticky" here? from what i see it just static on top always (which is default anyway)

frontend/src/components/ProtectedLayout/ProtectedLayout.styles.ts
```diff
@@ -0,0 +1,25 @@
...
+      'z-50',
+      'flex',
+      'justify-between',
+      'items-center',
+      'px-7.5',
+      'lg:px-12.5',
+      'py-4',
+      'bg-white',
+      'border-b',
+      'border-gray-200',
```
DariaAleshina: i dont see any borders in the figma file, and desktop has actually some greyish background colour (see the pic the event cards are actually white): 

Image

frontend/src/hooks/useAuth.ts
```diff
@@ -0,0 +1,57 @@
...
+}
+
+const MOCK_USER: UserData = { name: 'Fabio Rodrigues', initials: 'FR' };
+
+const useAuth = (): UseAuthReturn => {
+  const [user, setUser] = useState<UserData | null>(null);
+  const [loadingAuth, setLoadingAuth] = useState<boolean>(true);
+  const [authError, setAuthError] = useState<string | null>(null);
+
+  useEffect(() => {
```
DariaAleshina, line 23: are u sure we gonna do it with useEffect? (asking cause as i understand nowadays no one uses useEffect for fetching in real projects).

frontend/src/components/ProtectedLayout/ProtectedLayout.tsx
freeediver: Good catch @DariaAleshina ! The first time, I duplicated the logo to make it responsive using the two variants (full and compact). The problem with this approach is that it creates duplicate DOM elements, like this:
```tsx
 <div className="hidden md:block">
  <Logo size="full" />
</div>

<div className="md:hidden">
  <Logo size="compact" />
</div>
```

This works, but I’d like to ask if the Logo component itself should handle responsiveness. I suggest refactoring the rendering of the logo using CSS, like this:

```tsx
 export function Logo() {
  return (
    <>
      <span className="hidden md:inline">REDI. EVENTS.</span>
      <span className="md:hidden">R. E.</span>
    </>
  );
}
```

 I'd ask @fabi-rodrigues  which approach is better in this case.

frontend/src/components/ProtectedLayout/ProtectedLayout.tsx
freeediver: I'll try to implement the dedicated story and let you know.

frontend/src/components/ProtectedLayout/ProtectedLayout.styles.ts
freeediver: Thanks. I just updated it. You can check if it's better now.

frontend/src/hooks/useAuth.ts
freeediver: You're right! Since it wasn't specified in the ticket, I only created it provisionally. However, I agree that we should handle it differently. Perhaps we could use React Context and a custom useAuth() hook. Let's discuss it later.

frontend/src/components/UserArea/UserArea.tsx
```diff
@@ -49,7 +49,7 @@ const UserArea = ({ userName, avatarInitials, onProfile, onSignOut }: UserAreaPr
         aria-expanded={isExpanded}
       >
         <Avatar size="md" initials={avatarInitials} />
-        <div className="mx-3">{userName}</div>
+        <div className="mx-3 hidden md:inline">{userName}</div>
```
DariaAleshina: hey as you removed this div from mobile the margin horizontal is gone too (could you also fix it - i would personally make sure the parent div is a flex container with a gap-3 and removed margins at all): 

Image

frontend/src/hooks/useAuth.ts
DariaAleshina: maybe we keep it as s and then with next ticket properly connect to the backend.

frontend/src/components/ProtectedLayout/ProtectedLayout.stories.tsx
```diff
@@ -0,0 +1,30 @@
...
+};
+
+export const MobileViewport: Story = {
+  args: {
+    children: <div>Dashboard content</div>,
+  },
+  globals: {
+    viewport: { value: 'mobile2' },
+  },
+};
```
DariaAleshina, line 30: for me mobile story is not really mobile unless i change the viewport myself so maybe its just redundant 🤔. 

Image

frontend/src/components/UserArea/UserArea.tsx
freeediver: After taking a look at Figma, I realized that the UserArea component is missing some CSS properties, such as "gap," "height," and "padding." We won't need the margin for the username div anymore. I'll create a bug fix to address this issue.

frontend/src/components/ProtectedLayout/ProtectedLayout.stories.tsx
freeediver: I'll check it.

Antonov548 (approved):

PR #107, opened 2026-05-22
Readjust the UserArea styling based on the Figma CSS properties.

freeediver:
1- Adding missing properties inside the Header layout (userArea): h-[50px], px-[6px], gap-4
2- Correct colors: #F9F9F9 and #ECECEC
3- Update the Dropdown sizing: w-[136px], p-[15px], gap-[15px]


frontend/src/components/UserArea/UserArea.styles.ts
```diff
@@ -4,33 +4,36 @@ export const userArea = tv({
...
       'rounded-full',
+      'cursor-pointer',
       'transition-colors',
       'duration-300',
-      'hover:bg-gray-100',
-      'focus:outline-none',
-      'cursor-pointer',
+      'text-text-secondary',
+      'font-sans',
+      'text-base',
```
DariaAleshina: I would avoid adding font / text set up if its suppose to inherit it from main/parent set up (which is already font-sans and text-base (see global.css file).

frontend/src/components/UserArea/UserArea.styles.ts
```diff
@@ -4,33 +4,36 @@ export const userArea = tv({
...
       'flex',
-      'justify-between',
       'items-center',
-      'p-2',
+      'h-[50px]',
+      'px-[6px]',
+      'min-w-0',
+      'gap-2',
+      'sm:gap-4',
+      'bg-[#F9F9F9]',
```
DariaAleshina: we should not hard code colors, we use design tokens global css file (if your color is missing - just add it). 

but! I would argue specifically here that this component does not need its own default state background color as it will inherit it from protected layout (not critical tho).

frontend/src/components/UserArea/UserArea.styles.ts
```diff
@@ -42,15 +45,14 @@ export const userArea = tv({
       'border-none',
       'cursor-pointer',
       'hover:text-gray-700',
-      'focus:outline-none',
-      'focus:text-gray-700',
```
DariaAleshina, line 46: hey, why did you change it here? now the items are kinda inaccessible by keyboard (try navigate dropdown with a Tab, it does not visually highlight)

frontend/src/components/UserArea/UserArea.styles.ts
DariaAleshina: ahh ok now i see it was already there hehe

DariaAleshina (approved):

PR #108, opened 2026-05-24
Feat/issue 96/create user account

gokcesen:
Closes #96

Implements account creation for `POST /users`.

#### Changes

- Validates the user registration request
- Hashes the password before saving the user
- Creates new users with the `USER` role
- Prevents duplicate email registration
- Returns the created user without exposing `passwordHash`

### DIFF
--- a/backend/src/controllers/userController.ts
+++ b/backend/src/controllers/userController.ts
@@ -1,4 +1,4 @@
-import { Request, Response, NextFunction } from 'express';
+import { Request, Response } from 'express';
 import { UserService } from '../services/userService.js';
 
 const userService = new UserService();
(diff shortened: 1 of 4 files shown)


backend/src/controllers/userController.ts
```diff
@@ -20,11 +20,21 @@ export class UserController {
     res.json({ user });
   }
 
-  async createUser(req: Request, res: Response) {
-    const body = req.body;
-    const user = await userService.createUser(body);
+  async createUser(req: Request, res: Response, next: NextFunction) {
+    try {
```
akhilachiju: Since the project uses express-async-errors, the try/catch isn't needed here. Unhandled errors are automatically forwarded to the error middleware. You can simplify this to just call the service and return the response directly.

akhilachiju (commented):
.

backend/src/controllers/userController.ts
```diff
@@ -20,11 +20,21 @@ export class UserController {
...
+      const body = req.body;
+      const user = await userService.createUser(body);
+
+      return res.status(201).json({ user });
+    } catch (error) {
+      if (error instanceof Error && error.message === 'EMAIL_ALREADY_IN_USE') {
+        return res.status(400).json({
```
Davies70: Good job overall.

Small suggestion: `409 Conflict` may fit better than `400 Bad Request` for `EMAIL_ALREADY_IN_USE` since it’s a resource conflict rather than an invalid request.

MaksimTrofimov (approved):
Looks good to me!

PR #109, opened 2026-05-25
feat: exclude passwordHash from getAllUsers response

akhilachiju:
Resolves #97 

A user can request the list of users so the application can display user-related information where needed. The API returns public user data only.

What changed

Updated getAllUsers() in userService.ts to exclude passwordHash from every returned user using Prisma's omit option.

Why

To ensure the endpoint never exposes sensitive authentication data to the client.

Acceptance Criteria

- GET /users returns a list of users
- Each user includes id, firstName, lastName, email, and role
- Response does not include passwordHash
- Returns 200 with { users: [...] }
- Returns an empty array when no users exist


DariaAleshina (approved):
approve!

PR #111, opened 2026-05-26
implementing the signin in the frontend

santi-b-b:
Closes #110 
Closes #112 

Implemented sign in for the frontend
Managing errors and messages on the frontend. 
Implementing the middleware to protect private routes


frontend/src/hooks/useAuth.ts
```diff
@@ -0,0 +1,33 @@
+// src/hooks/useAuth.ts
+import { useState } from 'react';
+import { loginRequest } from 'src/service/authService';
+
+export function useAuth() {
+  const [loading, setLoading] = useState(false);
+  const [error, setError] = useState<string | undefined>(undefined);
```
FrontierPsychiatrist: You can write this easier like this

```typescript
const [error, setError] = useState<string>();
```

The react typings make `error` automatically have the type `string | undefined`.

frontend/src/hooks/useAuth.ts
```diff
@@ -0,0 +1,33 @@
...
+
+      return data; // token
+    } catch (err: unknown) {
+      if (err instanceof Error) {
+        setError(err.message);
+      } else {
+        setError('An unknown error occurred');
+      }
+
+      setTimeout(() => {
```
FrontierPsychiatrist: That really shouldn't be here. If the error should reset, IMO that should be done in another level of the application. This can also cause issues when unmounting anything that used the hook as the timeout is never cancelled. I'd suggest removing it completely for now.

frontend/src/hooks/useAuth.ts
```diff
@@ -0,0 +1,33 @@
+// src/hooks/useAuth.ts
+import { useState } from 'react';
+import { loginRequest } from 'src/service/authService';
+
+export function useAuth() {
+  const [loading, setLoading] = useState(false);
+  const [error, setError] = useState<string | undefined>(undefined);
+
+  const signIn = async (email: string, password: string) => {
```
FrontierPsychiatrist: It'd be good to have a return type documented in Typescript here.

frontend/src/service/authService.ts
```diff
@@ -0,0 +1,20 @@
+export async function loginRequest(email: string, password: string) {
```
FrontierPsychiatrist: It'd be good to have a return type documented in Typescript here. What's the shape of the token being returned? (Hint: It's `{ token: string }`)

frontend/src/app/(public)/sign-in/page.tsx
```diff
@@ -2,23 +2,25 @@
...
 import { SignInForm } from '@components';
+import { useAuth } from 'src/hooks/useAuth';
 
 export default function LoginPage() {
-  const handleSignIn = (data: { email: string; password: string }) => {
-    console.log('Form Ready for Backend:', data);
+  const { signIn, loading, error } = useAuth();
+
+  const handleSignIn = async (data: { email: string; password: string }) => {
+    await signIn(data.email, data.password);
```
FrontierPsychiatrist: This returns a token - but it's never being used anywhere. So while the login will kind of work here in the sense that credentials are exchanged for a token, subsequent requests that require the token will not be able to use it.

Basically, you will require to set up some application state that allows other requests to get the token for their requests.

frontend/src/hooks/useAuth.ts
```diff
@@ -0,0 +1,33 @@
+// src/hooks/useAuth.ts
```
FrontierPsychiatrist: I do not think that this is a useful comment.

FrontierPsychiatrist (commented):
This is a fantastic step for integrating frontend and backend.

Especially in terms of "logging in" there is one piece missing here though - after calling the login endpoint, the application is not really logged in since the returned token is just disregarded.
This could be handled in a subsequent PR. It's a quite drastic change to the application that has no easy solution really, sadly. Handling the authentication state can be quite complex. I am actually not sure what's currently recommended in the frontend to do that - maybe you can research a bit, or a teacher with more frontend experience can chime in.

Basically you want other requests to be able to set the `Authorization: Bearer ${token}` header after the login, but for that they _need_ the token first.

frontend/src/app/(public)/sign-in/page.tsx
```diff
@@ -2,23 +2,25 @@
...
 import Link from 'next/link';
 import { SignInForm } from '@components';
+import { useAuth } from 'src/hooks/useAuth';
 
 export default function LoginPage() {
-  const handleSignIn = (data: { email: string; password: string }) => {
-    console.log('Form Ready for Backend:', data);
+  const { signIn, loading, error } = useAuth();
+
+  const handleSignIn = async (data: { email: string; password: string }) => {
```
DariaAleshina: hi! i think we should also validate user input (unfortunately 🫠) with the ZOD schema (frontend/src/validators/schemas.ts) before submitting it. (taken this is urgent, we could skip it and create a new ticket for it, but should not forget about it).

frontend/src/service/authService.ts
```diff
@@ -0,0 +1,20 @@
...
+  const api_url = process.env.NEXT_PUBLIC_API_URL || 'http://localhost:4000';
+
+  console.log(`${api_url}/auth/login`);
```
DariaAleshina: do not forget to remove console logs before merging :)

frontend/src/app/(public)/sign-in/page.tsx
santi-b-b: I'll add it as moritz pointed out there's still some things to be implemented that I have to solve

frontend/src/service/authService.ts
santi-b-b: haha yeah I'll I was just debugging something that was not working at the time 😓

frontend/src/hooks/useAuth.ts
FrontierPsychiatrist: To extend a bit on this, if a component or a hook uses a timeout, you need to cancel it like this (and yes, it's ugly - maybe some frontend guru knows it better than me)

```typescript
const timeout = useRef();

useEffect(() => {
  return () => {
    if (timeout.current) {
      clearTimeout(timeout.current);
    }
  };
}, []);

function foo() {
  timeout.current = setTimeout(() => ..., xyz);
}
```

The `useRef` can be skipped if the timeout is being set in the effect itself of course. The core thing here is, that effects can return "cleanup" functions which will be called when they are exiting the scope/being unmounted.
This is especially important for anything that might try to update the DOM - React will log some angry errors if it's not being taken care of (though, in my experience, usually nothing _breaks_).

EDIT: And why `useRef` - it's a way to store values across re-renders whose value changes should _not_ trigger a re-render, that's why I used it over `useState` for the timeout here.

santi-b-b: I'll take it out for now I looks for an option to improve it after the merge : ) it was only a visual enhancement I came out with, but if it requires more complexity I don't think it should be blocking this

backend/src/controllers/authController.ts
```diff
@@ -12,7 +12,10 @@ export class AuthController {
         return res.status(400).json({ error: 'Email and password are required' });
       }
 
-      const token = await this.authService.login(email, password);
+      const data = await this.authService.login(email, password);
+      const token = data.token;
+      console.log('TOKEN FROM SERVICE:', token);
```
freeediver: Auth token said “keep me secret” but the console had other plans 👀
Probably best to remove these logs before prod.

frontend/src/utils/tokenStorage.ts
```diff
@@ -0,0 +1,15 @@
+const TOKEN_KEY = 'token';
+
+export const tokenStorage = {
+  set(token: string) {
+    localStorage.setItem(TOKEN_KEY, token);
```
FrontierPsychiatrist: I think this is a solution that's ok for this application, but just so you know - [OWASP](https://cheatsheetseries.owasp.org/cheatsheets/HTML5_Security_Cheat_Sheet.html#local-storage) does not recommend storing sensitive data in local storage, since any JS running on the domain has access to it. But mitigations are complex, so let's keep it - it's a good idea!

frontend/src/utils/tokenStorage.ts
```diff
@@ -0,0 +1,15 @@
...
+  },
+
+  get() {
```
FrontierPsychiatrist: Missing return type declaration in TypeScript. I _highly_ recommend _always_ declaring the return type, it leads to code where it's faster to detect errors using TypeScript. 

```typescript
get(): string
```

Imagine the following and let's assume you had a "contract" of `get` returning `string` in mind (never mind the obviously wrong implementation).

```typescript
function getToken() {
  return 1;
}

function test() {
  const token = getToken();
  token.substring(5);
}
```

Where do you see the compile error? Does it help you quickly (!) understand what is going wrong? Compare it to `function getToken(): string { return 1; }`.

backend/src/controllers/authController.ts
```diff
@@ -12,9 +12,17 @@ export class AuthController {
...
 
-      return res.status(200).json({ token });
+      res.cookie('token', token, {
+        httpOnly: true,
+        secure: process.env.NODE_ENV === 'production',
+        sameSite: 'lax',
+        maxAge: 1000 * 60 * 10,
+      });
+
+      return res.json({ ok: true });
```
santi-b-b, line 25: @FrontierPsychiatrist from the link you posted I think this could be a better way to store the token in a httpsonly cookie. how does it look to you?

FrontierPsychiatrist: I would not do that here. Yes - this is indeed one of the secure ways to "store" a token (though you already opened a can of worms with `lax` here, and they come with drawbacks, and I hope CORS is set up correctly), but:
1. This also changes the _transport_ of the token. [The authenticate middleware](https://github.com/ReDI-School/S26-Full-Stack-Circle/blob/main/backend/src/middleware/authenticate.ts) expects an `Authorization` header. This now switches to `Cookie`. You can of course adapt the middleware, but see 2)
2. To my knowledge it's more common to not use cookies these days. For the learning effect I think it's better if you keep the storage like you had it. I just mentioned OWASP so you can be aware for real-life projects, but for the learning purpose here I'd prefer not to have cookies. I _think_ (not sure) a secure variant in the browser uses service workers or something, but the FSC is not a security class, so let's not go to deep.

frontend/src/app/(private)/events/page.tsx
```diff
@@ -0,0 +1,30 @@
+import { config } from '@config';
+import { cookies } from 'next/headers';
+
+const EventsPage = async () => {
+  const cookieStore = await cookies();
+  const { apiUrl } = await config();
+  const token = cookieStore.get('token')?.value;
```
FrontierPsychiatrist, line 7: This is frontend, right? You can't access `httpOnly` cookies in JavaScript, this will always be `undefined`.

frontend/src/app/(private)/events/page.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+  const res = await fetch(`${apiUrl}/events`, {
+    method: 'GET',
+    headers: {
+      'Content-Type': 'application/json',
+      Authorization: `Bearer ${token}`,
```
FrontierPsychiatrist, line 13: If you switch to `httpOnly` cookies you do not need to set this header. The browser will automatically include the `Cookie` header with the cookie contained. For CORS requests you need to set `credentials: 'include'` on fetch, though.

frontend/src/app/(public)/sign-in/page.tsx
```diff
@@ -1,24 +1,33 @@
...
+import { LoginInput } from '@validators/schemas';
 
 export default function LoginPage() {
-  const handleSignIn = (data: { email: string; password: string }) => {
-    console.log('Form Ready for Backend:', data);
+  const { signIn, loading, error } = useAuth();
+  const router = useRouter();
+
+  const handleSignIn = async (data: LoginInput) => {
+    const token = await signIn(data.email, data.password);
```
FrontierPsychiatrist: `signIn` does not return a `token` anymore, just a boolean.

frontend/src/service/authService.ts
```diff
@@ -0,0 +1,20 @@
+import { config } from '../config';
+
+export async function loginRequest(email: string, password: string): Promise<void> {
+  const { apiUrl } = await config();
+
+  const res = await fetch(`${apiUrl}/auth/login`, {
+    method: 'POST',
+    credentials: 'include',
```
FrontierPsychiatrist, line 8: The login request doesn't need `credentials: 'include'`, at least I don't see why

frontend/src/middleware.ts
```diff
@@ -0,0 +1,38 @@
+import { jwtVerify } from 'jose';
+import { NextRequest, NextResponse } from 'next/server';
+
+const PUBLIC_AUTH_PATHS = ['/sign-in', '/sign-up'];
+
+async function isValidToken(token: string): Promise<boolean> {
+  const secret = process.env.JWT_SECRET;
```
FrontierPsychiatrist, line 8: The frontend can _NEVER_ (and this is crucially important) see the `JWT_SECRET`. That would immediately make the application unsafe. In the frontend, anyone can see the value and sign their own JWTs then, impersonating any other user.

I'm guessing this is frontend as it is in the `frontend` folder?

frontend/middleware.ts
```diff
@@ -0,0 +1,38 @@
...
+}
+
+export async function middleware(request: NextRequest) {
+  const { pathname } = request.nextUrl;
+  const token = request.cookies.get('token')?.value;
+
+  const isAuthPath = PUBLIC_AUTH_PATHS.some((path) => pathname.startsWith(path));
+  const authenticated = token ? await isValidToken(token) : false;
+
+  if (authenticated && !isAuthPath) {
```
FrontierPsychiatrist: That doesn't seem like a good idea, redirecting any request that is not the authentication? How is this meant to be used?

FrontierPsychiatrist (changes requested):
I understand the decision is to work with Cookies to store and transfer the authentication data which is fine.

This PR is in a weird limbo state now - the frontend reads from the cookies (despite them being `httpOnly`), it appears it also tries to read `JWT_SECRET`.

If you use cookies the frontend has much less work to do. It can't see the token and doesn't need to pass it in requests, this is then done by the browser.

I'm not sure I understand the middleware correctly, but the frontend must never (!!) see `JWT_SECRET`.

frontend/middleware.ts
santi-b-b: this is meant to take you to the home or events page if you're authenticated but you try to go to a public route such as the login or sign up page

frontend/src/middleware.ts
santi-b-b: Fabio actually coded this part, but this are my thoughts on it:

- this is technically frontend, but it runs in the server as it doesn't have the 'use client' definition at the top. So in this sense the secret won't be exposed to the browser. To me it looks like a safe implementation.

- Now I'm not sure if it would be "cleaner" to have this validation in the backend.

FrontierPsychiatrist: Ah ok, then it's indeed safe.

frontend/middleware.ts
FrontierPsychiatrist: But that's not what this checks, it checks if you are authenticated and _not_ on the sign or sign up page, if I understand correctly.

backend/src/controllers/authController.ts
FrontierPsychiatrist: As discussed on slack - a change to cookies is ok and wanted.

frontend/src/app/(private)/events/page.tsx
santi-b-b: this is only a mock page to test the authentication and redirection and its working. I'm not sure there's a point in putting too much work into it, as it is going to be replaced

frontend/src/app/(public)/sign-in/page.tsx
santi-b-b: that's true, now is a semantic thing token could be a boolean haha but is not the best name, I'll change it

frontend/src/service/authService.ts
santi-b-b: isn't it this include what sends the cookies to the backend to be checked? I took it out and the log in stopped work

FrontierPsychiatrist: Oh, I just double checked - it also controls if `Set-Cookie` in the response is respected. Didn't know that! Then it's needed, sorry for the noise.

Source: [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Request/credentials)

FrontierPsychiatrist (approved):

PR #122, opened 2026-06-03
feat: add cancel attendance endpoint

ssda-developer:
Close #55.

### DIFF
--- a/backend/src/controllers/attendanceController.ts
+++ b/backend/src/controllers/attendanceController.ts
@@ -1,4 +1,4 @@
-import { Request, Response } from 'express';
+import { NextFunction, Request, Response } from 'express';
 import { AttendanceService } from '../services/attendanceService.js';
 
 const attendanceService = new AttendanceService();
@@ -35,7 +35,7 @@ export class AttendanceController {
     }
   }
 
-  async cancelAttendance(req: Request, res: Response) {
+  async cancelAttendance(req: Request, res: Response, next: NextFunction) {
     try {
       const eventId = req.params.id;
       const userId = req.user!.userId;
@@ -48,7 +48,7 @@ export class AttendanceController {
         return res.status(404).json({ error: 'You are not registered for this event' });
       }
 
-      return res.status(500).json({ error: 'Internal server error' });
+      next(error);
     }
   }
 }
--- a/backend/src/controllers/attendanceController.ts
+++ b/backend/src/controllers/attendanceController.ts
@@ -34,4 +34,28 @@ export class AttendanceController {
       res.status(500).json({ error: 'Internal server error' });
     }
   }
+
+  async cancelAttendance(req: Request, res: Response) {
+    try {
+      const eventId = req.params.id;
+      const userId = req.user?.userId;
+
+      if (!userId) {
+        res.status(401).json({ error: 'Unauthorized' });
+        return;
+      }
+
+      await attendanceService.cancelAttendance(userId, eventId);
+
+      res.status(204).send();
+    } catch (error) {
+      if (error instanceof Error) {
+        if (error.message === 'NOT_REGISTERED') {
+          res.status(404).json({ error: 'You are not registered for this event' });
+          return;
+        }
+      }
+      res.status(500).json({ error: 'Internal server error' });
+    }
+  }
 }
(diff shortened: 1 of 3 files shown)


backend/src/controllers/attendanceController.ts
```diff
@@ -34,4 +34,28 @@ export class AttendanceController {
...
+
+      res.status(204).send();
+    } catch (error) {
+      if (error instanceof Error) {
+        if (error.message === 'NOT_REGISTERED') {
+          res.status(404).json({ error: 'You are not registered for this event' });
+          return;
+        }
+      }
+      res.status(500).json({ error: 'Internal server error' });
```
santi-b-b: @ssda-developer I think this is going to be handled by the middleware

backend/src/services/attendanceService.ts
```diff
@@ -38,4 +38,19 @@ export class AttendanceService {
...
     });
   }
+
+  async cancelAttendance(userId: string, eventId: string) {
+    const deleteResult = await prisma.attendance.deleteMany({
+      where: {
+        userId: userId,
+        eventId: eventId,
+      },
+    });
```
santi-b-b, line 48: why is this a delete many and not a simple delete?

backend/src/services/attendanceService.ts
```diff
@@ -38,4 +38,19 @@ export class AttendanceService {
...
+
+    if (deleteResult.count === 0) {
+      throw new Error('NOT_REGISTERED');
+    }
+
+    return deleteResult;
```
santi-b-b: you're returning the deleteResult, but not doing anything with it later.

backend/src/services/attendanceService.ts
ssda-developer: Because Prisma’s `delete` requires a unique ID, which means we'd have to do `findFirst` and then `delete` (2 queries). `deleteMany` allows filtering by `userId` and `eventId` directly, doing it in 1 query. Plus, it returns a count, which makes handling the 404 error easy.

backend/src/controllers/attendanceController.ts
ssda-developer: Good catch! You're totally right, the authenticate middleware already handles the 401 check.

backend/src/services/attendanceService.ts
santi-b-b: does it? from what I see here you can actually make the same filter, not an expert thought haha 

https://www.prisma.io/docs/orm/prisma-client/queries/crud#delete-a-single-record

backend/src/controllers/attendanceController.ts
```diff
@@ -34,4 +34,21 @@ export class AttendanceController {
...
+
+      await attendanceService.cancelAttendance(userId, eventId);
+
+      return res.status(204).send();
+    } catch (error) {
+      if (error instanceof Error && error.message === 'NOT_REGISTERED') {
+        return res.status(404).json({ error: 'You are not registered for this event' });
+      }
+
+      return res.status(500).json({ error: 'Internal server error' });
```
santi-b-b: the error handler middleware should take care of this one I think : )

backend/src/services/attendanceService.ts
santi-b-b: I mean attendances should be unique for a pair user-event, otherwise we have another kind of issue

backend/src/services/attendanceService.ts
ssda-developer: Fixed.

backend/src/services/attendanceService.ts
ssda-developer: You're totally right, we do have that unique index!

But if we use a simple .delete(), Prisma will throw a runtime error if the record isn't found. To handle that 404, we'd have to wrap the code in a messy try/catch and check for a specific Prisma error code (P2025).

.deleteMany() achieves the exact same thing in 1 query, but allows us to avoid try/catch completely by just checking deleteResult.count === 0. Since the DB index guarantees there's only 1 record anyway, it's 100% safe. Let's keep it for the sake of clean code?

backend/src/controllers/attendanceController.ts
ssda-developer: Fixed.

ssda-developer: Fixed.

backend/src/services/attendanceService.ts
santi-b-b: makes sense! good work : )

santi-b-b (approved):
Great job!

PR #123, opened 2026-06-05
feat: replace mock user with real data via server-side auth context

DariaAleshina:
- Introduces AuthContext seeded by the private layout server component. On every navigation into private routes, the layout reads the JWT cookie, decodes userId/role, fetches user data from the API – all server side – and passes it to AuthProvider.
- Removed mock data passed to ProtectedLayout and replaced with actual authenticated user from global context
- Profile page route (goToProfile navigates to /profiles/:id, placeholder until profile page is built)

### DIFF
--- a/frontend/src/app/(private)/layout.tsx
+++ b/frontend/src/app/(private)/layout.tsx
@@ -1,7 +1,31 @@
+import { cookies } from 'next/headers';
+import { decodeJwt } from 'jose';
 import { ProtectedLayout } from '@components';
+import { AuthProvider, AuthUser } from '@context/AuthContext';
+import { getUserById } from '@service/userService';
 
-const PrivateLayout = ({ children }: { children: React.ReactNode }) => {
-  return <ProtectedLayout>{children}</ProtectedLayout>;
+const PrivateLayout = async ({ children }: { children: React.ReactNode }) => {
+  let authUser: AuthUser | null = null;
+
+  try {
+    const token = (await cookies()).get('token')?.value;
+    if (!token) throw new Error('No token found');
+
+    const { userId, role } = decodeJwt<{ userId: string; role: 'USER' | 'ADMIN' }>(token);
+
+    const user = await getUserById(userId, token);
+    if (!user) throw new Error('Smth went wrong fetching user data');
+
+    authUser = { ...user, role };
+  } catch (err) {
+    console.error(err); // server-side only — logs to Node.js stdout, not the browser
+  }
+
+  return (
+    <AuthProvider initialUser={authUser}>
+      <ProtectedLayout>{children}</ProtectedLayout>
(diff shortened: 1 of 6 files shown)


frontend/src/app/(private)/layout.tsx
```diff
@@ -1,7 +1,31 @@
...
+
+    const { userId, role } = decodeJwt<{ userId: string; role: 'USER' | 'ADMIN' }>(token);
+
+    const user = await getUserById(userId, token);
+    if (!user) throw new Error('Smth went wrong fetching user data');
+
+    authUser = { ...user, role };
+  } catch (err) {
+    console.error(err); // server-side only — logs to Node.js stdout, not the browser
+  }
```
Antonov548, line 22: if feels like logic with receiving user and token should be inside AuthProvider. And `PrivateLayout` is just UI component which is child of `AuthProvider` and making sure that user is valid.

frontend/src/context/AuthContext.tsx
```diff
@@ -0,0 +1,32 @@
...
+interface AuthProviderProps {
+  initialUser: AuthUser | null;
+  children: React.ReactNode;
+}
+
+export const AuthProvider = ({ initialUser, children }: AuthProviderProps) => {
+  const [authUser, setAuthUser] = useState<AuthUser | null>(initialUser);
+
+  return <AuthContext.Provider value={{ authUser, setAuthUser }}>{children}</AuthContext.Provider>;
+};
```
Antonov548, line 35: I would maybe implement that AuthContext provide `user` field as readonly, so setter for it should be incampsulated (hidden) in `AuthProvider`, And in general interface of AuthProvider might looks like:
```ts
user: readonly object{}
login: (email, password) => {}
logout: () => {}
```

frontend/src/app/(private)/layout.tsx
DariaAleshina: The idea was to utilise Next.js capabilites. This is what i researched: In App Router, layouts are Server Components by default. A Server Component can be async, can call cookies(), and can make server-to-server API calls before sending anything to the browser. On the other hand AuthProvider is already 'use client' so it can't use cookies() (server-only) or be async. 

This is apparently the recommended pattern for seeding client state with server-fetched data. Correct me if i completely got it wrong 🥹

frontend/src/context/AuthContext.tsx
DariaAleshina: good point!

frontend/src/app/(private)/layout.tsx
Antonov548: Ah, okay. I see the intention, I'm not sure about it. 
Let's maybe also ask who worked more with Next.js specific features @fabiorodriguesdev

PR #126, opened 2026-06-05
create dashboard

Karenipo1:
…nd event filtering mismatch


Karenipo1:
I am trying to open a new Pull Request, because the previous was closed, related to 19-create-dashboard-page.

frontend/src/app/(private)/dashboard/page.tsx
```diff
@@ -0,0 +1,233 @@
...
+    const initializeDashboard = async () => {
+      await fetchDashboardEvents(isMounted);
+    };
+
+    initializeDashboard();
+
+    return () => {
+      isMounted = false;
+    };
+  }, [fetchDashboardEvents]);
```
santi-b-b: I think this might be to much logic for the page, I would consider maybe putting this in a custom hook? what do you think?

Also the mocks logics might not be the best idea to merge them into the main

frontend/src/service/authService.ts
```diff
@@ -18,11 +18,20 @@ export async function loginRequest(email: string, password: string): Promise<voi
...
+  const tokenValue = data.token || data.accessToken;
+
+  if (tokenValue) {
+    localStorage.setItem('token', tokenValue);
+  } else {
+    console.warn(
+      "Logueado, pero el backend no mandó un 'token' en el JSON. Estructura recibida:",
+      data
+    );
+  }*/
```
santi-b-b: Maybe is a good idea to take this out before merging

frontend/src/services/eventsService.ts
```diff
@@ -0,0 +1,137 @@
...
+        let filtered = MOCK_EVENTS;
+
+        if (tab === 'future') {
+          filtered = MOCK_EVENTS.filter((e) => new Date(e.date) >= today);
+        } else if (tab === 'archived') {
+          filtered = MOCK_EVENTS.filter((e) => new Date(e.date) < today);
+        }
+        setTimeout(() => resolve({ events: filtered }), 800);
+      });
+    }
```
santi-b-b: I wouldn't merge this logic into production

frontend/src/app/(private)/dashboard/page.tsx
Karenipo1: @santi-b-b  I've removed the redundant mock logic from the component as suggested. I did my absolute best to keep the code as clean as possible this time.

frontend/src/service/authService.ts
Karenipo1: done!

frontend/src/services/eventsService.ts
Karenipo1: @santi-b-b It was part of the ticket requirements, so I kept it. I'm not sure if we should remove it right now or later.

DariaAleshina (approved):

PR #128, opened 2026-06-05
refactor/auth-flow-enhancement

### LINKED ISSUE #127: Refactoring the Authentication workflow
Implementation details:

Implemented a full upgrade of the auth flow to cookie-based authentication with centralized client state management.

Backend

> Added cookie-parser and enabled it in index.ts.
> Updated authenticate.ts to read token from cookies first, fallback to Bearer token.
> Refactored login() to return { token, user }.
> Updated controller to return { ok: true, user } and added me() + logout() endpoints.
> Updated routes to support GET /me (protected) and POST /logout.

Frontend

> Extended auth service with AuthUser, getProfileRequest(), and logoutRequest().
> Introduced AuthContext to store and manage global auth state.
> Added AuthProvider wrapper and integrated it in layout.tsx.
> Rewrote useAuth to:
        - Use context state instead of local/mock state
        - Hydrate user on mount via /me
        - Handle sign-in/sign-out with API calls + redirects
>   Updated sign-up flow to use useAuth().signIn.
>   Fixed form typing for onSubmit.

Overall, the auth flow is now fully centralized, cookie-based, and shared across the app via context.

freeediver:
Implementation details:

Implemented a full upgrade of the auth flow to cookie-based authentication with centralized client state management.

Backend

1. Added cookie-parser and enabled it in index.ts.
2. Updated authenticate.ts to read token from cookies first, fallback to Bearer token.
3. Refactored login() to return { token, user }.
4. Updated controller to return { ok: true, user } and added me() + logout() endpoints.
5. Updated routes to support GET /me (protected) and POST /logout.

Frontend

1. Extended auth service with AuthUser, getProfileRequest(), and logoutRequest().
2. Introduced AuthContext to store and manage global auth state.
3. Added AuthProvider wrapper and integrated it in layout.tsx.
4. Rewrote useAuth to:

    - Use context state instead of local/mock state
    - Hydrate user on mount via /me
    - Handle sign-in/sign-out with API calls + redirects

5. Updated sign-up flow to use useAuth().signIn.
6. Fixed form typing for onSubmit.

Overall, the auth flow is now fully centralized, cookie-based, and shared across the app via context.

### DIFF
--- a/frontend/src/app/AuthProvider.tsx
+++ b/frontend/src/app/AuthProvider.tsx
@@ -1,0 +1,7 @@
+'use client';
+
+import { AuthProvider } from '@/contexts/AuthContext';
+
+export default function AuthProviderWrapper({ children }: { children: React.ReactNode }) {
+  return <AuthProvider>{children}</AuthProvider>;
+}
--- a/frontend/src/components/ProtectedLayout/ProtectedLayout.tsx
+++ b/frontend/src/components/ProtectedLayout/ProtectedLayout.tsx
@@ -4,10 +4,14 @@ import { Logo } from '../Logo';
 import { UserArea } from '@components';
 import useAuth from '@hooks/useAuth';
 import { Skeleton } from '@components/Skeleton';
+import { getInitials } from '@/utils/utils';
 
 export default function ProtectedLayout({ children }: ProtectedLayoutProps) {
   const { user, loading, signOut, goToProfile } = useAuth();
   const { main, headerBar, content } = protectedLayoutStyles();
+
+  const userFullName = user ? `${user.firstName} ${user.lastName}` : null;
+  const initials = user ? getInitials(userFullName!) : null;
 
   return (
     <main className={main()}>
@@ -28,8 +32,8 @@ export default function ProtectedLayout({ children }: ProtectedLayoutProps) {
           </div>
         ) : user ? (
           <UserArea
-            userName={user.name}
-            avatarInitials={user.initials}
+            userName={userFullName || ''}
+            avatarInitials={initials || ''}
             onProfile={goToProfile}
             onSignOut={signOut}
           />
--- a/frontend/src/components/ProtectedLayout/ProtectedLayout.tsx
+++ b/frontend/src/components/ProtectedLayout/ProtectedLayout.tsx
@@ -1,3 +1,5 @@
+import { useEffect } from 'react';
+import { useRouter } from 'next/navigation';
 import { protectedLayoutStyles } from './ProtectedLayout.styles';
 import type { ProtectedLayoutProps } from './ProtectedLayout.types';
 import { Logo } from '../Logo';
@@ -8,10 +10,15 @@ import { getInitials } from '@/utils/utils';
 
 export default function ProtectedLayout({ children }: ProtectedLayoutProps) {
   const { user, loading, signOut, goToProfile } = useAuth();
+  const router = useRouter();
+
+  useEffect(() => {
+    if (!loading && !user) router.replace('/sign-in');
+  }, [loading, user, router]);
   const { main, headerBar, content } = protectedLayoutStyles();
 
-  const userFullName = user ? `${user.firstName} ${user.lastName}` : null;
-  const initials = user ? getInitials(userFullName!) : null;
+  const userFullName = user ? `${user.firstName} ${user.lastName}` : '';
+  const initials = userFullName ? getInitials(userFullName) : '';
 
   return (
     <main className={main()}>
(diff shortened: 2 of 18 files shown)


frontend/src/hooks/useAuth.ts
```diff
@@ -1,44 +1,95 @@
...
+      }
+    };
+    hydrate();
+  }, [setAuthUser]);
 
-      await loginRequest(email, password);
+  // Clear stale error on mount
+  useEffect(() => {
+    setError(undefined);
+  }, []);
```
santi-b-b: Is this necessary? is causing the CICD to fail

frontend/src/app/AuthProvider.tsx
```diff
@@ -0,0 +1,7 @@
+'use client';
+
+import { AuthProvider } from '@/contexts/AuthContext';
+
+export default function AuthProviderWrapper({ children }: { children: React.ReactNode }) {
```
DariaAleshina, line 5: i have a feeling creating a wrapper in a separate file is kind of redundant, but maybe this is some convention i am not aware of ....

frontend/src/contexts/AuthContext.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+export const AuthProvider = ({ children }: { children: React.ReactNode }) => {
+  const [authUser, setAuthUser] = useState<AuthUser | null>(null);
+
+  const clearAuthUser = () => {
+    setAuthUser(null);
+  };
+
+  return (
+    <AuthContext.Provider value={{ authUser, setAuthUser, clearAuthUser }}>
```
DariaAleshina: this is the comment i received from a teacher in a similar PR i did - we should not expose the setter function - `setAuthUser` - , it should stay incapsulated. only specific actions like `clearAuthUser`

DariaAleshina: https://github.com/ReDI-School/S26-Full-Stack-Circle/pull/123/changes#r3363740080

frontend/src/hooks/useAuth.ts
```diff
@@ -1,44 +1,90 @@
...
+        }
+        return false;
+      } finally {
+        setLoading(false);
       }
+    },
+    [setAuthUser]
+  );
+
+  const signOut = useCallback(async () => {
```
DariaAleshina: are you sure useCallback is needed (here and in other cases here)? is there a performance issue? (we should avoid pre-mature performance optimisation).

frontend/src/hooks/useAuth.ts
```diff
@@ -1,44 +1,90 @@
...
-  };
+  }, [clearAuthUser, router]);
+
+  const goToProfile = useCallback(() => {
+    if (authUser) router.push(`/profiles/${authUser.id}`);
+  }, [authUser, router]);
 
-  return { signIn, loading, error, goToProfile, signOut, user, setUser };
+  return {
+    user: toUserData(authUser),
```
DariaAleshina: Are you sure we should limit user data here? 

My thoughts: ProtectedLayout needs full name & initials as i has specific UI needs but it is useless for other components... e.g. I assume very important will be to provide `userId` to check user relationship to events.  etc. I would provide all user data possible for potential UI needs, incl email and role. 

ProtectedLayout then will get user value (with firstName and lastName) from global state and make manipulations for its UI directly in the component. 

I guess the mock data was there in that shortened way initially just to test login with redirection to the draft "/events" page.

frontend/src/hooks/useAuth.ts
```diff
@@ -1,44 +1,90 @@
...
+      // Continue clearing local state
     } finally {
-      setLoading(false);
+      clearAuthUser();
+      router.push('/sign-in');
     }
-  };
+  }, [clearAuthUser, router]);
+
+  const goToProfile = useCallback(() => {
```
DariaAleshina: lets remove useCallbacks, i don't think it is needed, and just makes code harder to comprehend and maintain.

frontend/src/contexts/AuthContext.tsx
freeediver: Update done, thx

frontend/src/hooks/useAuth.ts
freeediver: You're right, UseCallback doesn't help here. All consumers re-render, regardless of the methods.

frontend/src/components/ProtectedLayout/ProtectedLayout.tsx
```diff
@@ -4,11 +4,15 @@ import { Logo } from '../Logo';
...
 import { Skeleton } from '@components/Skeleton';
+import { getInitials } from '@/utils/utils';
 
 export default function ProtectedLayout({ children }: ProtectedLayoutProps) {
   const { user, loading, signOut, goToProfile } = useAuth();
   const { main, headerBar, content } = protectedLayoutStyles();
 
+  const userFullName = user ? `${user.firstName} ${user.lastName}` : null;
+  const initials = user ? getInitials(userFullName!) : null;
+
```
DariaAleshina: hm maybe assign empty string to userFullName and initials directly here instead of on line 35-36? also i guess initials should depend on userFullName ? ... not just the user hehe.

frontend/src/components/ProtectedLayout/ProtectedLayout.tsx
DariaAleshina: should Skeleton also trigger if (!user) but it wont in current logic . it wull get to null..meaning nothing will render at all.  not sure this is exactly the behavior we want for the protected layout. what do u think?

DariaAleshina (approved):
Hi, Reda! Great massive work! Made some fixes, tho I did not have chance to review it all really well. Yet it works and we need global state asap, so I approve with comments for further follow up:

- (not urgent) Implement server side user hydration (/me call) as now it is client side each page load and there is a visible delay. 
- Vercel deployment preview its still not working :( lets investigate further if we can fix it by demo day.

PR #129, opened 2026-06-08
feat: add archived state to EventCard component and update stories

Davies70:
- closes #118
- Implemented archived state for EventCard
- Added StoryBook test

### DIFF
--- a/frontend/src/components/EventCard/EventCard.tsx
+++ b/frontend/src/components/EventCard/EventCard.tsx
@@ -31,7 +31,8 @@ export default function EventCard(props: EventCardProps) {
   } = EventCardStyles();
 
   const buttonVariant = variantMap[props.action];
-  const buttonState = props.action === 'archived' ? 'disabled' : 'default';
+  const buttonState =
+    props.actionState ?? (props.action === 'archived' ? 'disabled' : 'default');
 
   const timeStamp = formatTimestamp(props.date);
 
@@ -56,7 +57,12 @@ export default function EventCard(props: EventCardProps) {
               {props.attendeeCount} of {props.maxAttendees}
             </p>
           </div>
-          <Button variant={buttonVariant} state={buttonState} size="small" onClick={props.onActionClick}>
+          <Button
+            variant={buttonVariant}
+            state={buttonState}
+            size="small"
+            onClick={props.onActionClick}
+          >
             <span className={buttonText()}>{props.action}</span>
           </Button>
         </div>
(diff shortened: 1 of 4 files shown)


frontend/src/components/EventCard/EventCard.tsx
```diff
@@ -30,6 +31,7 @@ export default function EventCard(props: EventCardProps) {
   } = EventCardStyles();
 
   const buttonVariant = variantMap[props.action];
+  const buttonState = props.action === 'archived' ? 'disabled' : 'default';
```
DariaAleshina: hey hey! you added button state which makes sense, and as the button could receive "loading" state theoretically (e.g. (when API is called on clicking "join" or "leave") but this logic here will give only 2 alternatives always: "default" or "disabled". Also here it is a derived value from action prop, tho in types file (line 63) it seems like you are adding it as a stand alone prop to be passed to the EvenCard but then this logic here kinda kills it...

DariaAleshina: let me know what you think, this is the only place where I got confused.  Overall great that you initiated adding this state :) we really need it

Davies70: Thanks for your review. I added the loading state. Let me know if you have more questions.

frontend/src/components/EventCard/EventCard.types.ts
```diff
@@ -55,6 +56,17 @@ type EventCardLoadedProps = EventCardBaseProps & {
...
+  /**
+   * Controls the action button UI state.
+   *
+   * - `default` — standard button behavior
+   * - `loading` — shows the button loading indicator and disables click
+   * - `disabled` — disables the button
+   *
+   * This can be managed by the parent when an API call is in progress.
+   */
+  actionState?: 'default' | 'disabled' | 'loading';
```
DariaAleshina: i would suggest to change this prop design cause we only need "loading" here ("disabled "depends on "archived" action and can be derived). I would change it to a more clear boolean type `isActionPending?: boolean;`.

frontend/src/components/EventCard/EventCard.tsx
```diff
@@ -30,6 +31,8 @@ export default function EventCard(props: EventCardProps) {
   } = EventCardStyles();
 
   const buttonVariant = variantMap[props.action];
+  const buttonState =
+    props.actionState ?? (props.action === 'archived' ? 'disabled' : 'default');
```
DariaAleshina: The current fallback logic lets actionState override the archived → disabled rule. (imagine if smw passed both "actionState: loading" and "action: Archived" at same time)

Suggest flipping the priority and simplifying the prop to a boolean:
```
const buttonState = props.action === 'archived'
  ? 'disabled'
  : props.isActionPending ? 'loading' : 'default';
```

Davies70: Hi Daria, I have made your recommended changes.

DariaAleshina (approved):
Great job!🔥

PR #130, opened 2026-06-08
feat: add user settings page with name and password update

akhilachiju:
Resolves #114

A logged-in user can update their display name and password from the settings page. The form integrates with the existing backend user service via httpOnly cookies for authentication.

What changed

- Added UserSettingsForm component with first name, last name, and password fields — pre-filled with the logged-in user's current name
- Added settings/page.tsx and settings/layout.tsx under the private route
- Added Settings entry to the UserArea dropdown with a GearSixIcon
- Added updateUserSchema and UpdateUserInput to validators/schemas.ts, reusing the shared passwordSchema already used in registration
- Update logic is handled locally in the settings page (not in useAuth)
- Auth context is updated on save so the header name reflects changes immediately
- Form skips the API call if nothing has changed
- Removed tokenStorage — authentication is handled exclusively via httpOnly cookies
- Added translations for the Settings menu entry (en.json, de.json)

Why

The app already uses httpOnly cookies for authentication set on login. No token storage in localStorage is needed or appropriate — the browser sends the cookie automatically on every request. Password 
validation reuses the existing passwordSchema to stay consistent with registration rules.

Acceptance Criteria

- PUT /users/:id updates the user's first name and last name
- Password is updated only when a new password is provided
- Leaving the password fields blank keeps the existing password
- New password must meet the same strength requirements as registration
- Response does not expose passwordHash
- Settings page shows a success message on save
- Settings page shows an error message if the request fails
- First name and last name are pre-filled with the current user's values and persist after refresh
- No API call is made if the user submits without making any changes
- Header name updates immediately after saving without a page refresh


frontend/src/utils/tokenStorage.ts
```diff
@@ -0,0 +1,26 @@
...
+    const token = this.get();
+    if (!token) return null;
+    try {
+      const payload = JSON.parse(atob(token.split('.')[1]));
+      return payload.userId ?? null;
+    } catch {
+      return null;
+    }
+  },
+};
```
santi-b-b: I think we are using a cookies approach to authentication, aren't we?

akhilachiju: Yes, exactly! We're using httpOnly cookies. I also removed a leftover tokenStorage.set(data.token) line that was still in authService.ts from an old token-based approach — that's cleaned up now.

backend/src/services/userService.ts
```diff
@@ -39,10 +39,21 @@ export class UserService {
     });
   }
 
-  async updateUser(id: string, data: { email?: string; firstName?: string; lastName?: string }) {
+  async updateUser(id: string, data: { email?: string; firstName?: string; lastName?: string; password?: string }) {
```
FrontierPsychiatrist: Avoid inline types like this when they get to large, it is better to extract a type for this. It's hard to read.

frontend/src/hooks/useAuth.ts
```diff
@@ -40,6 +42,26 @@ export default function useAuth() {
     }
   };
 
+  const updateUser = async ({ firstName, lastName, newPassword }: UpdateUserInput) => {
```
FrontierPsychiatrist: That is no `authentication` issue and should not be in this hook. It's also wrongly reusing state from the `auth` hook. Please move this to a different hook, or just keep it local to where it is needed (I don't assume the user will be updated all over the application?

frontend/src/services/authService.ts
```diff
@@ -13,9 +13,7 @@ export async function loginRequest(data: LoginInput): Promise<AuthUser> {
   const res = await fetch('/api/auth/login', {
     method: 'POST',
     credentials: 'include',
-    headers: {
-      'Content-Type': 'application/json',
-    },
+    headers: { 'Content-Type': 'application/json' },
```
FrontierPsychiatrist, line 16: Please avoid unnecessary formatting changes in your PRs. This is not related to editing the user.

frontend/src/utils/tokenStorage.ts
```diff
@@ -0,0 +1,26 @@
+const TOKEN_KEY = 'token';
+
+export const tokenStorage = {
```
FrontierPsychiatrist: Please remove this complete file, it is not related to the user editing and a substantial add to the authentication system.

frontend/src/utils/tokenStorage.ts
FrontierPsychiatrist: If you're using cookies you do not need token storage, cookies are already stored by the browser.

FrontierPsychiatrist (changes requested):
The token storage _must_ be removed. You should not just casually store sensitive credentials in local storage in a PR about editing user data, especially when there is already an authentication system (cookies) in place.

The `updateUser function must not be in the `auth` hook, it is no authentication feature.

frontend/src/utils/tokenStorage.ts
akhilachiju: Done — tokenStorage.ts has been deleted entirely. We're using httpOnly cookies exclusively now, no local storage involved.

backend/src/services/userService.ts
akhilachiju: Done.

frontend/src/services/authService.ts
akhilachiju: Noted.

frontend/src/hooks/useAuth.ts
akhilachiju: Done — updateUser has been moved out of useAuth and inlined directly into the settings page, since it's only used there. No separate hook needed.

frontend/src/utils/tokenStorage.ts
akhilachiju: Done.

frontend/src/components/UserSettingsForm/UserSettingsForm.tsx
```diff
@@ -0,0 +1,71 @@
...
+  });
+
+  return (
+    <form onSubmit={handleSubmit(onSubmit)} className="flex flex-col gap-6 w-full" noValidate>
+      {serverError && <InfoBox variant="error" message={serverError} />}
+      {successMessage && <InfoBox variant="success" message={successMessage} />}
+
+      <div className="flex flex-col gap-4">
+        <InputField
+          label="First name"
```
wms198, line 39: Hi, I think that could be nice to have currently first name as default value. 
Imaging that the user only want to change the password, then its unnecessary to tip first name and last name again.

Image

akhilachiju: Good point! I've added default values for first name and last name from the logged-in user, so they're pre-filled in the form. Only the password fields stay empty.

PR #131, opened 2026-06-08
125 fix layout shift on 404 network error pages

DariaAleshina:
- closes #125 

- Add flex-1 to the content wrapper on /network-error and /not-found so the layout fills the full height without shifting
- Move /network-error out of PUBLIC_AUTH_PATHS into a separate ALWAYS_PUBLIC_PATHS bypass in middleware, so the page is reachable by both authenticated and unauthenticated users (previously authenticated users were redirected to /events)

TEST by:

- visiting [/network-error](http://localhost:3000/network-error) - should open when both authenticated and not. 
- visiting any non-existent route (for 404 page)

### DIFF
--- a/frontend/src/app/(public)/network-error/NetworkErrorView.tsx
+++ b/frontend/src/app/(public)/network-error/NetworkErrorView.tsx
@@ -11,7 +11,7 @@ const NetworkErrorView = () => {
   };
 
   return (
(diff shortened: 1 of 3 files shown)


Davies70 (commented):
Good work, Daria. I know this isn’t your ticket, but is the general sign-in implementation working on your end? After signup or sign-in, does it transition to the Events page?

Davies70 (approved):
Looks good to me.

PR #133, opened 2026-06-08
feat: add localization with language switcher in user profile menu

### LINKED ISSUE #113: Localization
- [ ] Add aditional checkbox for the language in user profile menu
- [ ] Add localization for the app using multiple (decide on the supported) languages
- [ ] Library for the localization

Image

akhilachiju:
Resolves #113

A logged-in user can switch the app language from the user profile menu. The selected language persists across sessions via localStorage.

What changed

- Added next-intl for localization
- Added messages/en.json and messages/de.json with English and German translations for the UserArea strings (Profile, Sign Out)
- Added IntlProvider component wrapping the root layout — manages locale state, persists to localStorage, and provides NextIntlClientProvider and a useLocale context hook
- Added src/i18n/locales.ts defining supported locales (en, de), default locale, and display names
- Added LanguageSwitcher component rendered inside the UserArea dropdown — shows current language with a translate icon, expands to reveal other available languages on click
- Updated UserArea to use useTranslations for Profile and Sign Out labels

Why

The app had no localization support. Users should be able to use the app in their preferred language. The switcher lives in the user profile menu for easy access.

Acceptance Criteria

- Language switcher appears in the user profile dropdown
- Default language is English
- Clicking the current language reveals the other available language(s)
- Selecting a language switches all translated strings immediately
- Selected language persists across page refreshes via localStorage
- Adding new languages only requires a new entry in locales.ts and a new messages/xx.json file


frontend/messages/de.json
```diff
@@ -0,0 +1,11 @@
+{
+  "userArea": {
+    "profile": "Profil",
+    "signOut": "Abmelden",
+    "language": "Sprache"
+  },
+  "languages": {
+    "en": "Englisch",
```
FrontierPsychiatrist: Shouldn't the language always be in it's native tongue? Otherwise it's hard for someone to switch to their language if they end up on the page in a language they don't speak. Probably not an issue for an English speaker to understand that "Englisch" is "English", but in general that might not be the case (e.g. I would not be able to figure out on a completely Chinese page to switch to German if it just displayed 德文 - hope Google translated that correct).

frontend/src/components/UserArea/UserArea.tsx
```diff
@@ -42,9 +51,7 @@ const UserArea = ({ userName, avatarInitials, onProfile, onSignOut }: UserAreaPr
         ref={buttonRef}
         type="button"
         className={styles.userArea({ isExpanded })}
-        onClick={() => {
-          setIsExpanded((prev) => !prev);
-        }}
+        onClick={() => setIsExpanded((prev) => !prev)}
```
FrontierPsychiatrist: Unnecessary formatting change

frontend/src/components/UserArea/UserArea.tsx
```diff
@@ -62,9 +69,40 @@ const UserArea = ({ userName, avatarInitials, onProfile, onSignOut }: UserAreaPr
...
+          {/* Language switcher */}
+          <button
+            type="button"
+            className={styles.item()}
+            onClick={() => setIsLangExpanded((prev) => !prev)}
+          >
+            <span>{tLang(locale)}</span>
+            <TranslateIcon className={styles.icon()} aria-hidden="true" focusable="false" />
+          </button>
+          {isLangExpanded && (
```
FrontierPsychiatrist: Extract components to avoid having too large components that are hard to understand. This whole div could be its own component.

FrontierPsychiatrist (changes requested):
Please do not translate language names in other languages as it makes switching hard for people who do not understand multiple languages.

You could try to extract a component for the language switcher to not bloat the user area component.

For future, not now: When doing i18n, backend requests should also set the [`Accept-Language`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Accept-Language) header so the backend can return correctly localized data. Since we don't do that now, it's not necessary to add.

Try to avoid adding formatting changes to your PR that are unrelated to the functional changes, as it puts more burden on the reviewer to filter out what changes are actually necessary to review.

frontend/messages/de.json
akhilachiju: Fixed — language names are always displayed in their own language (English, Deutsch) regardless of the active locale, sourced from localeNames in locales.ts rather than the message files.

frontend/src/components/UserArea/UserArea.tsx
akhilachiju: Reverted — restored to the original multi-line onClick form.

frontend/src/components/UserArea/UserArea.tsx
akhilachiju: Extracted into src/components/LanguageSwitcher/LanguageSwitcher.tsx. UserArea now just renders <LanguageSwitcher />.

FrontierPsychiatrist (dismissed):

FrontierPsychiatrist (approved):

PR #138, opened 2026-06-09
feat: implement delete own user account

Tatianatod:
Implemented DELETE /users/:id route and checks.


santi-b-b (approved):
Looks good! well done : )

PR #139, opened 2026-06-09
Feat/issue 121/navigate to home page

akhilachiju:
Resolves #121

A user visiting the app is redirected to the correct page based on their auth state. Clicking the logo in the protected layout navigates back to the home page.

What changed

- Updated frontend/src/app/page.tsx to check for an auth token and redirect to /events if signed in, or /sign-in if not (removes the previous TODO)
- Wrapped the logo image in a Link in Logo.tsx so clicking it navigates to /

Why

The home page had a hardcoded redirect to /sign-in with a TODO comment. Users who are already signed in should land on the dashboard. The logo in the protected layout had no navigation, so users had no way 
to return home by clicking it.

Acceptance Criteria

- Visiting / while signed in redirects to /events
- Visiting / while not signed in redirects to /sign-in
- Clicking the logo in the protected layout navigates to /


frontend/src/components/Logo/Logo.tsx
```diff
@@ -36,14 +37,15 @@ const LOGO_VARIANTS_DICTIONARY: LogoVariantsDictionary = {
...
   return (
-    <Image
-      src={LOGO_VARIANTS_DICTIONARY[size].textColor[textColor]}
-      alt="ReDi Events"
-      aria-hidden="true"
-      width={LOGO_VARIANTS_DICTIONARY[size].width}
-      height={LOGO_VARIANTS_DICTIONARY[size].height}
-      priority
-    />
+    <Link href="/">
```
DariaAleshina: Suggestion (non-blocking): I'd move the <Link> wrapper out of this file so Logo stays reusable in non-navigation contexts — not in scope right now, but plausible later (footer, loading states). The link wrapping can live in the layout/header, or in a standalone LogoLink component if it's needed in multiple places.

I'd also add an aria-label to make the link's purpose explicit for screen reader users:
```
<Link href="/" aria-label="Go to homepage">
  <Logo ... />
</Link>
```
CTA like "Go to homepage" describes the action, whereas the image alt ("ReDi Events") only names the brand. Note: with the label on the link, the alt is ignored for the link's name, so alt="" inside this usage avoids redundancy.

frontend/src/app/page.tsx
```diff
@@ -1,6 +1,20 @@
...
+
+const Home = () => {
+  const router = useRouter();
+  const { user, loading } = useAuth();
+
+  useEffect(() => {
+    if (!loading) {
+      router.replace(user ? '/events' : '/sign-in');
+    }
+  }, [user, loading, router]);
```
DariaAleshina: Nice! 🙂 Not sure about this one - should this maybe be handled by middleware instead? We already have middleware.ts in place doing rerouting based on auth, so the / → /events redirect could potentially live there too. Might save us the client-side roundtrip (i guess blank page while useAuth resolves). Non-blocking though from my side

frontend/src/components/Logo/Logo.tsx
akhilachiju: Done! Added aria-label="Go to homepage" to the link and set alt="" on the image to avoid redundancy. Agree on moving the Link wrapper out of Logo in a follow-up — noted for later.

frontend/src/app/page.tsx
akhilachiju: Good call! Moved the / → /events redirect into middleware so it's handled server-side. Cleaned up page.tsx as well — no more client-side roundtrip.

frontend/src/components/Logo/Logo.types.ts
```diff
@@ -10,4 +10,6 @@ export interface LogoProps {
    * @default "default"
    */
   textColor?: 'default' | 'white';
+
```
FrontierPsychiatrist, line 13: Unnecessary whitespace change

FrontierPsychiatrist (approved):

frontend/src/components/Logo/Logo.types.ts
akhilachiju: Noted

PR #140, opened 2026-06-09
delete account

wms198:
I added a deleteMe endpoint, otherwise I would need to build something else to get the current users id.

Open for feedback!

[Delete account](https://github.com/ReDI-School/S26-Full-Stack-Circle/issues/115) 

image
image


frontend/src/components/ProfileCard/ProfileCard.tsx
```diff
@@ -11,6 +13,15 @@ const ProfileCard = ({
   participatedEvents,
 }: ProfileCardProps) => {
   const nameInitials = getInitials(name);
+  const delteProfile = async () => {
+    try {
+      await deleteMe();
+      alert('User deleted');
```
wms198: Hi @santi-b-b , I added a native browser alert for confirmation of deletion which is also a requirement of my ticket.

santi-b-b: aah ok sorry I missed it. now correct if I'm wrong but on story book the button doesn't look like its clickable, it doesn't have the cursor type nor a hover style

backend/src/controllers/userController.ts
```diff
@@ -55,4 +55,15 @@ export class UserController {
     await userService.deleteUser(id);
     return res.status(204).send();
   }
+
+  async deleteMe(req: Request, res: Response) {
+    const authenticatedUser = req.user;
+
+    if (!authenticatedUser) {
```
FrontierPsychiatrist: This is part of the `authentication` middleware. do not do that check again. My recommendation, use `req.user!` in this case.

backend/src/controllers/userController.ts
```diff
@@ -55,4 +55,15 @@ export class UserController {
...
+      return res.status(401).json({ error: 'Unauthorized' });
+    }
+
+    await userService.deleteUser(authenticatedUser.userId);
+    return res.status(204).send();
```
FrontierPsychiatrist, line 66: I don't think `send` is needed.

frontend/src/components/ProfileCard/ProfileCard.tsx
FrontierPsychiatrist: That is just an alert _after_ the user has been deleted. Instead a `window.confirm` should be used _before_ the delete.

FrontierPsychiatrist (changes requested):
Remove duplicated user check from the controller, use `window.confirm` _before_ deleting instead of `alert` _after_.

PR #144, opened 2026-06-11
feat/22 create event edit page

### LINKED ISSUE #21: Create Event Creation Page
### Description

Image

- https://www.figma.com/design/itAzcC43MCVpSyJCcmQJgu/ReDi-Events?node-id=0-1

Create the `EventCreationPage` using `ProtectedLayout`. Allows the user to create a new event with an empty form.

### Acceptance Criteria

- [ ] Uses `ProtectedLayout`
- [ ] Sub-header: `LinkButton` "← GO BACK" on the left only (no "CREATE NEW EVENT" button since user is already creating)
- [ ] `FormCard` with title "CREATE NEW EVENT" containing:
  - [ ] `InputField` Title* — empty
  - [ ] Row with three fields: Date* (date), Time* (time), Capacity* (number) — empty
  - [ ] `InputField` Description* — textarea, empty
  - [ ] Footer: "CANCEL" button (gray) + "CREATE EVENT" button (teal)
- [ ] No `DangerCard` on this page
- [ ] **Validation:** all fields required, show field-level errors on submit
- [ ] "CANCEL" navigates back to dashboard
- [ ] "CREATE EVENT" submits the form (mock) and navigates to the new event details page
- [ ] **Responsive (mobile):**
  - [ ] Single column layout, form takes full width
  - [ ] Date / Time / Capacity row stays inline on mobile (compact inputs)
  - [ ] "GO BACK" link remains at the top

Aarbei:
Closes #21


frontend/src/app/(private)/events/[id]/edit/EditEventFormClient.tsx
```diff
@@ -0,0 +1,231 @@
+'use client';
+
+import { useState } from 'react';
+import { useRouter } from 'next/navigation';
+import Link from 'next/link';
+import { LinkButton } from '@components/LinkButton';
```
Comment (santi-b-b) [inline frontend/src/app/(private)/events/[id]/edit/EditEventFormClient.tsx:6]: Hey! is this a component? should it go in the components folder?

frontend/src/app/(private)/events/[id]/edit/page.tsx
```diff
@@ -0,0 +1,150 @@
...
+      return { ok: false, error: 'Event id is missing' };
+    }
+
+    const { apiUrl: actionApiUrl } = await config();
+    const actionCookieStore = await cookies();
+    const actionToken = actionCookieStore.get('token')?.value;
+
+    if (!actionToken) {
+      return { ok: false, error: 'No token provided. Please log in.' };
+    }
```
Comment (santi-b-b) [inline frontend/src/app/(private)/events/[id]/edit/page.tsx:67]: I think this should be already handled by the middleware. If this is a protected route, the middleware should allow only authenticated users to get to this page

backend/prisma/seed.ts
```diff
@@ -76,7 +76,7 @@ async function main() {
       id: 'c1d8e3f6-4b72-4a95-7d01-9e2f4c6b8a3d',
       title: 'Demo day',
       description: 'Meeting for everyone to demonstrate projects',
-      date: new Date('2026-05-17'),
+      date: new Date('2026-05-17T10:00:00.000Z'),
```
FrontierPsychiatrist, line 79: Why was this change necessary? It is important to explain this in commit messages or the PR description for the reviewer.

backend/src/controllers/eventController.ts
```diff
@@ -7,6 +7,10 @@ const eventService = new EventService();
 const attendanceService = new AttendanceService();
 type EventFilter = 'upcoming' | 'past';
 
+function composeEventDateTime(date: string, time?: string): Date {
```
FrontierPsychiatrist: `new Date()` will use the system time zone to parse the date which especially on a server will be UTC. So when a request comes in with the time `13:00` from let's say Germany, it's meant to be CEST at the moment. This will be parsed as `11:00` CEST though.

---

At some point it is better to rely on a date time library for this (sadly the JS standard date time is rather bad and [Temporal](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Temporal) is too new.

Examples are luxon, date-fns, moment.

But apart from this, this function is missing documentation. From the implementation I can assume the caller is _meant_ to pass in a date string in the `yyyy-MM-dd` format and an optional time string in `HH:mm` format, but that is impossible to know from the function name. Please add `JSDoc` properly explaining how to use function.

backend/src/controllers/eventController.ts
```diff
@@ -121,10 +125,18 @@ export class EventController {
       });
     }
 
+    let composedDate: Date | undefined;
+
+    if (req.body.date || req.body.time) {
+      const datePart = req.body.date ?? event.date.toISOString().slice(0, 10);
+      const timePart = req.body.time ?? event.date.toISOString().slice(11, 16);
```
FrontierPsychiatrist: This can be a bad idea as you might lose the correct timezone here. `toISOString` will return a UTC timestamp, but you cut that off. Then you re-parse the date in the system timezone with `composeEventDateTime`. Currently this means, at least in CEST timezone this will constantly change the time two hours back of the event (if the request does not specify a time).

frontend/src/app/(private)/events/[id]/edit/EditEventFormClient.tsx
```diff
@@ -0,0 +1,231 @@
...
+type Props = {
+  eventId?: string;
+  eventHref: string;
+  initialEvent: EditEventFormData;
+  onUpdateEvent: (payload: UpdateEventPayload) => Promise<EventActionResult>;
+  onDeleteEvent: () => Promise<EventActionResult>;
+};
+
+export default function EditEventFormClient({
+  eventId,
```
Comment (FrontierPsychiatrist) [inline frontend/src/app/(private)/events/[id]/edit/EditEventFormClient.tsx:22]: (This is a personal taste comment): I know that most people (especially in frontend) do _not_ do this, but I do not recommend destructuring, especially large objects. Why?

1. It takes away vertical screen space. We work on widescreen monitors, vertical screen estate is valuable. This now "wastes" 5 lines for basically nothing.
2. It removes some nice scoping when reading the component code. E.g. when I see `props.eventId` I immediately can understand that the `eventId` is coming from the properties. In this component, when I see `eventId` I have to trace back to where it was defined.

Again, I am not asking for a change here, I wanted to comment on this to give my reasoning, that's all. So for clarity, I would use

```typescript
export default function EditEventFormClient(props: Props) {...
```

FrontierPsychiatrist (changes requested):
Only required change: The timezone handling when updating an event is incorrect when the server has a different timezone than the client due to the reliance on `new Date()`'s default timezone.

frontend/src/app/(private)/events/[id]/edit/EditEventFormClient.tsx
```diff
@@ -0,0 +1,231 @@
...
+            GO BACK
+          </LinkButton>
+        </div>
+        <div className="hidden lg:inline-flex">
+          <Link href="/events">
+            <Button variant="idle">CREATE NEW EVENT</Button>
+          </Link>
+        </div>
+        <div className="lg:hidden">
+          <Link href="/events">
```
Comment (gokcesen) [inline frontend/src/app/(private)/events/[id]/edit/EditEventFormClient.tsx]: The "CREATE NEW EVENT" button currently links to `/events`, which is the events dashboard/list page from my understanding. 

Based on the existing create-event flow, this might need to navigate to `/create-event` instead.

backend/src/controllers/eventController.ts
santi-b-b: I'm not sure if it is the same case but in the file utils/formatTimestamp there's a function we use for the eventCard

const timePart = new Intl.DateTimeFormat('en-US', {
    hour: 'numeric',
    minute: '2-digit',
    hour12: true,
  }).format(date);

  return `${datePart} – ${timePart}`;
};

santi-b-b (approved):
Overall, the code looks good and the feature works in general terms.

However, there are a few structural and architectural points worth improving for consistency with the rest of the codebase:

- The implementation does not fully follow the established pattern (hook / service / BFF / presentation-only components).  In several places, logic is handled directly in components or controllers instead of being centralized in services.
- There is some duplication of logic, particularly around event retrieval and related helper functions, which could be reused more consistently through shared services.
- The introduction of timezone handling feels a bit premature in this scope. It adds noticeable complexity across frontend and backend without a proportional impact on user-facing behavior at this stage.
- There are a few missing or incomplete styles that affect visual consistency with the rest of the UI.
- The delete event flow is not fully robust yet; there are some missing pieces that can lead to inconsistent behavior or failure states.

Overall, it’s a good implementation, but it would benefit from some refactoring to align better with the existing architecture and to simplify the current timezone handling.

backend/prisma/seed.ts
santi-b-b: this was done in order to store the time zone within the date

backend/src/services/eventDateService.ts
```diff
@@ -0,0 +1,81 @@
+import { DateTime } from 'luxon';
+
+export class EventDateService {
```
FrontierPsychiatrist, line 3: (Style only) There's no real need for this to be a class. It doesn't encapsulate any state. It might as well be exported functions. But not a requested change!

backend/src/services/eventDateService.ts
```diff
@@ -0,0 +1,81 @@
...
+  private readonly DEFAULT_TIME_ZONE = 'Europe/Berlin';
+
+  /**
+   * Local (form input) → UTC Date (DB)
+   */
+  localToUtc(params: { date: string; time?: string; timezone: string }): Date {
```
FrontierPsychiatrist, line 9: JSDoc should document expected format of date and time. It's impossible to use this function:

```typescript
localToUtc({ date: 'Yesterday', time: 'around noon', timezone: 'UTC' })
```

FrontierPsychiatrist (approved):
I don't have capacity to fully understand the TZ code now, but at a glance it looks correct.

PR #146, opened 2026-06-11
Fix/sign in form validation

freeediver:
Some technical optimization:  
1- create two separate contexts  inside the provider 
-  useAuthContext (Public) : { authUser, isHydrating } - read-only, no mutation
- useAuthDispatch (Internal/ just useAuth hook which is calling it): { setAuthUser }-  used only by useAuth.ts

2- Hydration now runs exactly once app-wide inside the Provider.


DariaAleshina (approved):
🔥

PR #147, opened 2026-06-12
Fix create event sumbit and links

DariaAleshina:
- fix: proxy POST /events through Next.js API route to fix Vercel preview

Direct client fetch used a build-time env var for the backend URL and
relied on cross-domain cookies (SameSite=Lax), both of which break on
Vercel preview deployments. Adds /api/events server route that resolves
the backend URL at request time and forwards the auth token from the
httpOnly cookie as an Authorization header.

- fix: route name for redirect to event creation page from Event page


santi-b-b (approved):
it works! amazing

PR #150, opened 2026-06-12
Add user event filtering and retrieval by user ID

Davies70:
- closes #120 

## getEventsByUserId

Adds a new service method, controller handler, and route to fetch events filtered by user involvement.

### What changed
- `eventService.ts` — added `getEventsByUserId(userId, filter)` with date-aware filtering for `created`, `attending`, and `archived` states
- `eventController.ts` — added `getEventsByUserId` handler with filter validation and auth guard
- `eventRouter.ts` — added `GET /user/events?filter=` route, placed before `/:id` to avoid param collision

### Usage
GET /events/user/events?filter=created    → future events the user organised

GET /events/user/events?filter=attending  → future events the user is attending

GET /events/user/events?filter=archived   → past events (organised or attended)


backend/src/controllers/eventController.ts
```diff
@@ -142,4 +152,23 @@ export class EventController {
 
     return res.status(200).json({ event: updatedEvent });
   }
+
+  async getEventsByUserId(req: Request, res: Response) {
+    const userId = req.user?.userId;
+
+    if (!userId) {
```
FrontierPsychiatrist: This should be done by the `authenticate` middleware and not be duplicated in the controller function (see `eventRoutes.ts`).

I understand that this is probably a problem with `user` being `| undefined` in the request and TypeScript complaining, but this is duplicating logic all over the codebase _and_ potentially the error codes are different.

This is a pretty tricky one to get around, as no type definition of middlewares can guarantee that the user exist. I'd actually say that the `!` escape hatch is valid here, even though it must be used sparingly overall:

```typescript
const userId = req.user!.userId
```

Yes, this can lead to errors if the `authenticate` middleware is not used - but that's a pretty tough error that would cause substantial errors anyway and basically must be caught by tests.

backend/src/services/eventService.ts
```diff
@@ -102,4 +102,23 @@ export class EventService {
       data,
     });
   }
+
+  async getEventsByUserId(userId: string, filter: UserEventFilter) {
+    const now = new Date();
+
+    const where: Prisma.EventWhereInput =
+      filter === 'created'
```
FrontierPsychiatrist: Overusing the ternary operator is very hard to read here. My recommendation: Never nest ternaries.

You could use a `switch` here, even though it has the small disadvantage of not being an expression in TS/JS:

```typescript
let where: Prisma.EventWhereInput;
switch(filter) {
  case 'created':
    where = { organizerId: userId, date: { gte: now } };
    break;
  case 'attending':
    where = { attendances: { some: { userId } }, date: { gte: now } };
    break;
  default:
    where = {
      date: { lt: now },
        OR: [{ organizerId: userId }, { attendances: { some: { userId } } }],
      };
}
```

backend/src/services/eventService.ts
```diff
@@ -102,4 +102,23 @@ export class EventService {
       data,
     });
   }
+
+  async getEventsByUserId(userId: string, filter: UserEventFilter) {
+    const now = new Date();
```
FrontierPsychiatrist: In general not a good idea for testability to use "now" in service-level functions, it's better to pass it as a parameter.

FrontierPsychiatrist (commented):
Overall looks pretty good.

Please remove the `userId` check from the controller as it is duplicating functionality that is part of the `authenticate` middleware. The rest are recommendations.

backend/src/services/eventService.ts
```diff
@@ -102,4 +102,27 @@ export class EventService {
       data,
     });
   }
+
+  async getEventsByUserId(userId: string, filter: UserEventFilter, now: Date = new Date()) {
+    let where: Prisma.EventWhereInput;
+
+    switch (filter) {
```
MaksimTrofimov, line 142: If default is used, then any new or incorrect filter value will automatically fall into it. This can lead to incorrect results — for example, instead of throwing an error, the user may receive archived events even if the request is invalid.

For this reason, default should not be used as a fallback option. Instead, an explicit exception should be thrown so that errors are detected immediately rather than being converted into incorrect data.

```
switch (filter) {
  case 'created':
  case 'attending':
  case 'archived':
    break;

  default:
    throw new Error(`Unknown filter: ${filter}`);
}
```

backend/src/controllers/eventController.ts
```diff
@@ -30,6 +30,16 @@ function parseEventFilter(value: unknown): {
   };
 }
 
+function parseUserEventFilter(value: unknown): {
+  isValid: boolean;
+  filter?: UserEventFilter;
+} {
+  if (value === 'created' || value === 'attending' || value === 'archived') {
```
MaksimTrofimov: It’s better to use constants because the values are not duplicated across the codebase.
If a new filter is added or an existing one is changed, you only need to update it in one place.
This reduces errors (like typos) and makes the code easier to maintain.

backend/src/services/eventService.ts
Tatianatod: Done. Thank you, Maksim.

backend/src/controllers/eventController.ts
Tatianatod: Also done. Thanks!

freeediver (approved):

PR #153, opened 2026-06-15
fix: event creation form display mobile

DariaAleshina:
- closes #151 

Potentially problem with mobile devises related. used appearance-none for mobile to reset any browser specific styling of an input field. thanx @ssda-developer for the tip!

Final result on my device (iPhone): 
IMG_3959


ssda-developer (approved):
Great job!

PR #154, opened 2026-06-16
fix: event time display

DariaAleshina:
all new events (created from the form) would fall to time **2AM**. Issue was that eventController would skip "time" input at all and not recorded in a database. Fixed now!


santi-b-b (dismissed):
great finding!!

santi-b-b (approved):
lovely

